PR #72, opened 2016-12-11
procman - Process Management for AzuraCast

Xe:
WIP.

This will do the following:
- [ ] Start binary with exec arguments, job name, healthcheck kind
  - [ ] If it dies respawn it
  - [ ] Heartbeat monitoring: call a healthcheck function once a second, store its results in 
  - [ ] If the process fails healthchecks 3 times out of the last 5 tries, it is restarted and azuracast is alerted by some api path as of yet to be determined
- [ ] HTTP API to control jobs (probably use port 2600 and json message passing)

### DIFF
--- a/util/procman/README.md
+++ b/util/procman/README.md
@@ -1,0 +1,4 @@
+procman
+=======
+
+Like a megaman robot master except he doesn't have free will and won't turn against you (I hope)


SlvrEagle23:
Prospective location for shared configuration file could be `/var/azuracast/www_tmp/stations.json`

File format could be similar to:

```
{
  'base_dir': '/var/azuracast/stations',
  'stations': [
    {
      'id': 1,
      'short': 'azuratest_radio'
      'frontend_type': 'icecast',
      'frontend_config': {},
      'backend_type': 'liquidsoap',
      'backend_config': {},
    }
  ]
}
```

Based on this config, it can then assume management roles for the processes associated with the backend (i.e. LiquidSoap) and frontend (i.e. IceCast/ShoutCast).

It could then expose an API for various functions:
* Start process
* Stop process
* Is process running
etc

PR #327, opened 2017-12-12
Update default.j2

susl16c:
Can I leave it like this?


SlvrEagle23:
@susl16c This number is far too high to be acceptable in common use. You can set up your own instance to have those limits, but I find the existing 50MB limit to be more than permissible for the purposes of uploading files, especially since the media uploads are separated into even smaller chunks.

PR #334, opened 2017-12-13
Please update the translation

susl16c:
I have done so many days


SlvrEagle23:
@susl16c As I've mentioned to you before, all translation work is done via CrowdIn. Please update the translations used on this page: https://crwd.in/azuracast

CrowdIn generates the language files that we then use in AzuraCast. Please don't directly pull request these files.

PR #461, opened 2018-02-28
Custom JS in public page

Nontenda:
Hi,

I've added a field in the Branding config page to write javascript and display it on public radio page.
I've test it successfully.

If someone may review this, I'm of course willing to work again on that to improve this code.

### DIFF
--- a/app/config/forms/branding.conf.php
+++ b/app/config/forms/branding.conf.php
@@ -37,6 +37,16 @@ return [
             ]
         ],
 
+        'custom_js_public' => [
+            'textarea',
+            [
+                'label' => _('Custom JS for Public Pages'),
+                'description' => _('This javascript code will be applied to the station public pages and login page.'),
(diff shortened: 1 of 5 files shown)


Nontenda:
Please note that I'm not familiar with the files used for translation, I've updated the lines in default.pot but I'm not sure if it will affect the other translations.

SlvrEagle23:
@Nontenda Really nice work on this. Always happy to see a new pull request, especially one that catches things I myself even forget (like locale updates).

Just in case you were editing that .pot file by hand, for future reference there's a command you can run to do it automatically, `./util/docker-compose-dev.sh run --rm cli locale_generate`

I would be more scrutinizing of the security implications of including custom JS, but considering this is only something global system administrators can apply, and they have to manually add it to a form specifically for this purpose, I'm not so worried. Merging!

PR #610, opened 2018-06-08
Custom stream url field

terokorp:
Admin can set custom urls to stream when using proxy or authentication or other special cases.
Needs translations.

### DIFF
--- a/app/config/forms/mount/icecast.conf.php
+++ b/app/config/forms/mount/icecast.conf.php
@@ -97,6 +97,14 @@ return [
                     ]
                 ],
 
+                'custom_listenurl' => [
+                    'text',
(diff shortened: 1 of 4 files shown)


SlvrEagle23:
@terokorp Thank you very much for your contribution! Sorry for the delay in getting it merged in, I'm dealing with some medical stuff and my laptop's out of commission, so I can't process things with my normal speed.

Everything looks great in the PR, though I would change the variable to `custom_listen_url` and update the function names to say `getCustomListenUrl` for consistency. If you get a chance and can make those changes, I can merge directly into the main branch, otherwise I'll merge it into a feature branch, make the minor update and then roll it upstream.

Issue #898, opened 2018-10-26
Deactivate streamer account for configurable time after manual disconnect


Vaalyn:
**Is your feature request related to a problem? Please describe.**
When you disconnect a streamer via the station profile site the streamer gets disconnected but popular streaming software like for example VirtualDJ will automatically reconnect after the manual disconnect. This makes the disconnect streamer function not quite as reliable for kicking a dj off of a stream when he/she is doing something that needs to be stopped since he/she will just reconnect without any real delay or anything.

**Describe the solution you'd like**
I'd like to be able to configure an amount of time that the streamer account of a streamer that is manually disconnected will be deactivated for so that a manual disconnect forces a hard limit on when a dj can reconnect after he/she is thrown off the stream.

Depending on how granular the time should be configurable there needs to be some kind of cronjob that checks if the time that a streamer account should be deactivated has passed and activate the account if true. If it should only be configurable in minutes then an addition to the `syncShort` sync tasks would be enough. If it should be configurable in seconds than there needs to be a new kind of sync task that runs approximately every second.

The time should be configurable in the edit station profile screen.

To keep track of when a streamer account can be reactivated the easiest way would be to add a field to the `StationStreamer` entity called `reactivate_at` of type `int | nullable` where we set the timestamp of the moment where the account is allowed to be reactivated when the disconnect streamer function is executed. When the cronjob runs we get every `StationStreamer` that has a value for `reactivate_at` and check if the current timestamp is equal or higher than the value of `reactivate_at`, activate the account and set the `reactivate_at` back to null.

**Describe alternatives you've considered**
I can't really think of any good alternatives besides having a not changeable time limit for this but this feature would be kind of annoying if you can't configure it like you individually need it.

**Additional context**
Related to issue #670

SlvrEagle23:
@Vaalyn Yep, this was pretty much exactly what I had in mind for dealing with the "disconnect streamer" issue not having any sort of meaningful permanence for streaming apps that automatically reconnect streamers.

I may go the approach of not necessarily having it be a static configuration but rather a dropdown on the "Disconnect Streamer" button that says "...for 5 minutes, ...for 1 hour, disable account" etc

Vaalyn:
@SlvrEagle23 I'll send you a PR some time today with an implementation of the deactivation of the streamer after disconnecting and reactivation after the time passed. 👍

NeddyPlays:
[...]

PR #900, opened 2018-10-26
Deactivate streamer on manual disconnect - implements #898

Vaalyn:
This PR implements the feature requested in issue #898

The time that a streamer is deactivated for can be configured in seconds so that this feature is as flexible as possible but the `Task` that checks if a streamer can be reactivated will run in the 15s TaskRunner so that this check does not cost too much performance. This could be changed to a dedicated TaskRunner that executes every second to give a kind of realtime precision for this feature but I'm not sure if this is really necessary.

The dropdown mentioned by @SlvrEagle23 in this issue can be implemented rather easily on top of these changes. The `disconnectStreamer()` method parameters would need to be extended by an optional `int $seconds` parameter. If this parameter is set use this to determine the `reactivate_at` timestamp if not use the default from the station. The disconnect dropdown would then need to send the amount of seconds with the `POST` request to the backend action. You could define some pre-configured values for this dropdown and set the station default value as the default option of the dropdown. This would enable station owners to define a custom value to primarily use and also adds some easy to use options.

### DIFF
--- a/config/forms/station.php
+++ b/config/forms/station.php
@@ -284,6 +284,17 @@ return [
                         'step' => '0.1',
                     ]
                 ],
+                
+                'disconnect_deactivate_streamer' => [
+                    'number',
+                    [
+                        'label' => __('Deactivate Streamer on Disconnect (Seconds)'),
+                        'description' => __('Number of seconds to deactivate station streamer on manual disconnect. Set to 0 to disable deactivation completely.'),
+                        'default' => 0,
+                        'min' => '0',
+                        'step' => '1',
+                    ]
+                ],
 
                 'use_manual_autodj' => [
                     'radio',
--- a/src/Entity/Migration/Version20181025232600.php
+++ b/src/Entity/Migration/Version20181025232600.php
@@ -1,0 +1,28 @@
+<?php declare(strict_types=1);
+
+namespace App\Entity\Migration;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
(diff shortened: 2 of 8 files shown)


SlvrEagle23:
@Vaalyn :o

This is...this is some absolutely excellent code. I think this may be the first PR I've received that directly affects the AzuraCast core to this degree and yet demonstrates so significant of an understanding of both the code itself and modern PHP best practices.

I'm looking through everything and it all looks...exactly right. I can't find anything I'd modify before merging this into the core.

Very nice work on this, and thank you for the contribution. I'll play with this on local a little bit and then get it merged into the core ASAP.

Vaalyn:
> @Vaalyn :o
> 
> This is...this is some absolutely excellent code. I think this may be the first PR I've received that directly affects the AzuraCast core to this degree and yet demonstrates so significant of an understanding of both the code itself and modern PHP best practices.
> 
> I'm looking through everything and it all looks...exactly right. I can't find anything I'd modify before merging this into the core.
> 
> Very nice work on this, and thank you for the contribution. I'll play with this on local a little bit and then get it merged into the core ASAP.

Thanks =)

I try to stay on top of the current best practices and to match the style and architecture of the projects I work on to keep the code bases consistent. The AzuraCast code and structure is easy to follow. Had no problems finding how everything works together.

I like this project and I may contribute some more features in the future. There is already another issue I've identified while playing with AzuraCast that impacts the scheduling of playlists and I also have some performance issues when adding songs to a playlist but I'm still trying to identify where the performance issue is coming from. Will create issues for this later and try to implement / fix these.

Issue #740, opened 2018-08-22
Moving audio files


aljaxus:
**Is your feature request related to a problem? Please describe.**
I'd like to be able to move music files in `/stations/ID/files`

**Describe the solution you'd like**
To be able to move files by dragging them

**Describe alternatives you've considered**
An alternative solution would be a separate menu / popup / modal that would open, instead of being able to simply drag the file

LazosPlaying:
// talking about moving them into / out of folders

tropimega:
no puedo agregar mas musica, a que se debe, espero respuestas gracias

aljaxus:
> no puedo agregar mas musica, a que se debe, espero respuestas gracias

what ?

Issue #575, opened 2018-04-29
Possibility to move music files between maps


RemBdev:
Hello,

I've a feature request: The possibility to move music files between maps.

Thanks for the great product!

Greetings.

Issue #315, opened 2017-12-06
Moving files from folders


susl16c:
Can I move files from one folder to another?

SlvrEagle23:
@susl16c This isn't currently possible from the web file manager. You can always connect via SFTP and manually manipulate the files: https://github.com/AzuraCast/AzuraCast/blob/master/SUPPORT.md#access-files-via-sftp-docker-installations

PR #958, opened 2018-11-16
Implement audio file moving

Vaalyn:
This PR implements a basic file moving function for the audio files. Currently this feature can only move files and not directories but it should be possible to extend this in the future to add the moving of whole directories.

~~I've also already rebuilt the static assets like it's shown in the documentation.~~ <- Wasn't needed

Should close the following issues: #740, #315 and maybe #575 (I don't really understand what "maps" means in that issue so I assume it means directories)

### DIFF
--- a/config/routes.php
+++ b/config/routes.php
@@ -333,6 +333,9 @@ return function(\Slim\App $app)
             $this->map(['GET', 'POST'], '/list', Controller\Stations\Files\FilesController::class.':listAction')
                 ->setName('stations:files:list');
 
+            $this->map(['GET', 'POST'], '/directories', Controller\Stations\Files\FilesController::class.':listDirectoriesAction')
+                ->setName('stations:files:directories');
+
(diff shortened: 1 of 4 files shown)


SlvrEagle23:
@Vaalyn It doesn't look like you changed any of the assets built in the `/web/static` directory, so there's no need to rebuild the static assets. The JS file you updated is added inline to the file management page, so it's not bundled in with the static assets.

Vaalyn:
@SlvrEagle23 thanks for pointing that out. I don't know how I didn't notice that 🤔

I've reversed that commit now.

SlvrEagle23:
@Vaalyn Excellent work on this. Everything uses the existing architecture to accomplish what it needs to do, and does so in about as elegant of a way as is possible given our software stack.

I still eventually want to, at some point, just rip out the file manager and rebuild it as something far more "drag and drop" and interactive, akin to a file explorer, but this is an excellent addition to what we have now.

Thank you again for your contribution!

Issue #1003, opened 2018-11-30
Stream URLs on station profile page missing


Vaalyn:
**Installation method**
Docker

**Host OS (for Docker installations)**
Debian 9

**Describe the bug**
This commit b1a3473 causes the stream URLs to not be shown anymore on the station profile page.

**Expected behavior**
The stream URLs should be shown on the station profile page.

**Screenshots**
http://prntscr.com/loy5uq

**Additional context**
This happens because the `getCustomListenUrl` method of the `StationMount` entity now returns an empty string instead of a `null` which causes [this condition](https://github.com/AzuraCast/AzuraCast/blob/b1a34737d923a8e7a48368186a28454adb5f9b2c/src/Radio/Frontend/FrontendAbstract.php#L134) to resolve to true thus returning an empty URI.

PR #1004, opened 2018-11-30
Fix condition check for custom listen url - fixes issue #1003

Vaalyn:
This fixes issue #1003


SlvrEagle23:
@Vaalyn Preferred to use `!empty` in this case because the value could either be the empty string or NULL if it was never set. Implemented in another commit direct to master.

PR #1015, opened 2018-12-02
Automatic update Azuracast with docker at a specific time and remove old files

RonVerburg:
Hi because er are a lot of updates every day is there a way to automatic update Azuracast with docker at a specific time and remove old files.
Example every day at 01.00 hour a automatic update en after that remove old files to keep the systeem clear.

Is this a request thats posible to do?

R.v.B


SlvrEagle23:
@RonVerburg This definitely should be created as an issue, not a pull request. See this URL for creating an issue: https://github.com/AzuraCast/AzuraCast/issues/new/choose

PR #1062, opened 2018-12-23
Fixes for file playtime detection

TraceOrdin:
StationMediaRepository.php:  sometimes getID3 will report an error in a file
due to a slightly mangled header but can still get an accurate playtime length
and some other info. In the loadFromFile function removed the condition that
there are no errors prior to checking playtime so that we can still get that info,
and instead sent any detected errors to the php log.

Flow.php: In _createFileFromChunks function changed final filename to use actual
filename uploaded instead of putting _complete at the end so that getID3 can
use the file extension as a fallback identifier if the standard mp3 header is
slightly mangled. That fallback occurs in GetFileFormat() in
/var/azuracast/www/vendor/james-heinrich/getid3/getid3/getid3.php

Note: there might be a better way to do the logging, I went the way I'm familiar with using php's error_log but this might be better done with Monolog

### DIFF
--- a/src/Entity/Repository/StationMediaRepository.php
+++ b/src/Entity/Repository/StationMediaRepository.php
@@ -178,38 +178,46 @@ class StationMediaRepository extends Repository
     {
         // Load metadata from supported files.
         $id3 = new \getID3();
-
+ 
         $id3->option_md5_data = true;
         $id3->option_md5_data_source = true;
         $id3->encoding = 'UTF-8';
 
         $file_info = $id3->analyze($file_path);
 
-        if (empty($file_info['error'])) {
+        // Report any errors found by the file analysis to the logs
+        if (!empty($file_info['error'])) {
+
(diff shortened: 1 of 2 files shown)


SlvrEagle23:
@TraceOrdin Thank you for your contribution. I apologize for the delay in getting back to it; we've been out for the holidays and also dealing with some issues with our static assets.

I will review the changes and get back to you shortly!

TraceOrdin:
Glad to help! A bunch of files I was uploading were showing 0 for the time even though getID3 was pulling the times OK in a different environment. In my case the specific error getID3 was finding is: 'Invalid MPEG audio header (ff e9 eb 7a ) at offset 74', but the files play ok and the calculated length is accurate :)

SlvrEagle23:
@TraceOrdin Approved and merged; I'll go back here shortly and roll the logging for that section into Monolog.

Issue #368, opened 2018-01-09
Define size quota for each radio station


multi023:
Hi,

One question: Is there any way to assign auto-dj size to each radio station?

On the control panel of each radio it appears: "13.31 GB of 19.10 GB Used (5.78 GB Free)". On that radio, I've uploaded about 70 songs, which will be about 250 MB. In each radio there are 13 GB occupied?

Is there any way to define the disk size for each auto-dj?

If someone can help me...thank you thousands.

SlvrEagle23:
@multi023 At the moment, this isn't possible for one primary reason: because our main recommended method of distributing AzuraCast is via Docker, which (by design) abstracts out things like the host's CPU, RAM and hard drive storage so that the individual containers don't need to particularly worry about managing them.

On some Docker hosts, it's possible to set quotas for each of these items, but the support varies widely and isn't something we could include directly in the distributed Docker Compose file without causing a lot of potential issues. I went looking for a relevant Docker Compose configuration section in case you wanted to add it in just for your specific instance, but the documentation on this is so sparse that I can't even find _that_.

As for creating a fake quota that we only enforce when attempting to upload files directly through the web interface, that's possible, though it could easily be circumvented by anyone uploading using SFTP directly to the filesystem.

xewl:
> As for creating a fake quota that we only enforce when attempting to upload files directly through the web interface, that's possible, though it could easily be circumvented by anyone uploading using SFTP directly to the filesystem

I don't see that as a problem, as long as the application warns you, and denies uploads (unless admin, or permission). A limit could be tracked by checking the Station's data folder (recursively), perhaps with a cached cronjob to lessen the load, instead of iterating on every request (maybe remove cached value on each upload).

multi023:
Hi,

I would be happy to use instead of 20 GB, for example, to use 50 or 100 GB. Is that if not, I can not create more radio stations on a 2 TB server, having used only 20 GB :(

![datos de nuevo servidor](https://user-images.githubusercontent.com/20141463/36982077-2ffd2f1e-208f-11e8-9c5a-02cc29f0501e.jpg)


It's great, it does not consume resources ... well, I'm still happy, but I could be much happier :)

SlvrEagle23:
@multi023 The problem you're having isn't related to this issue, which is requesting the _addition_ of a soft quota per-station into the system. No such quota exists right now.

[...]

PR #1122, opened 2019-01-28
Implement #368 -- Add administrator-set storage quotas to individual stations.

SlvrEagle23:


williambatson:
I just updated to check out the new quote system - I have 4 stations-  all 4 reporting 16gb used of the 10gb allocated.... including one station with nothing/no data.   What did I do wrong?
https://gyazo.com/2ba180807227954bc960fc6cb58f6aa1

SlvrEagle23:
@williambatson Give it until the first 5-minute sync runs, or just manually run the 5-minute sync from the console administration. The default "used" size is the disk space used on the partition otherwise.

williambatson:
hmmm -  mins says hasnt run in 24 mins, i clciked run task and got:
500
The given value "" does not represent a valid number.
BigNumber.php : 67

Now for sure I think I broke it!

SlvrEagle23:
@williambatson Found and fixed an error relating to the new code; update and let me know if the sync task runs properly.

williambatson:
Mostly right now -  ran the update, the stations with files already on them report correctly on the 5min update and now show properly on the music files page.
the only one with trouble is the one with 0 songs/data.. shows:
total_size
"0 (0.0 B)"
total_files
0
unchanged
0
updated
0
created
0
deleted
0
I have it set to 10gb -  files page still shows 16gb of 10gb used and will not allow uploads at all - uploads denied. - Again - the others with files already loaded seem fine now.

SlvrEagle23:
@williambatson That one was caused by a bit of PHP fun (the string '0' is `empty` to PHP).

I've updated the code to not use the partition filled size if no file size is available.

Thanks for working with me on resolving this!

PR #1127, opened 2019-01-30
Fix issues found with PHPStan

Vaalyn:
With the help of [PHPStan](https://github.com/phpstan/phpstan) I've found several issues that I've fixed in this PR.  I have used PHPStan with the option `--level 1` (and also excluded some errors that will be resolved through correct config of PHPStan) so there are most likely more issues that would be shown on higher levels.

I'll create an issue to discuss if we want to add PHPStan as a dev dependency, on which level we want to check the codebase against and if this should be run automatically on PRs.

### DIFF
--- a/src/Controller/Api/Admin/RolesController.php
+++ b/src/Controller/Api/Admin/RolesController.php
@@ -122,7 +122,7 @@ class RolesController extends AbstractGenericCrudController
 
                         if (!empty($value['station'])) {
                             foreach($value['station'] as $station_id => $station_perms) {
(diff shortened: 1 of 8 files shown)


SlvrEagle23:
@Vaalyn The `test` route is actually used by me in dev and needs to stick around. Otherwise, looks fine!

SlvrEagle23:
@Vaalyn Actually, scratch that...I'm gonna do something like `routes.dev.php` that loads only if you're in a dev environment and is entirely gitignored, to avoid this being a potential issue.

PR #1505, opened 2019-05-04
Add libsamplerate to Ansible installation

Vaalyn:


SlvrEagle23:
@Vaalyn don't forget to update `update.sh` to increment the revision number

Vaalyn:
Thanks for the reminder 👍

PR #1588, opened 2019-05-25
Add environment variable to make the amount of php-workers configurable

Vaalyn:


SlvrEagle23:
@Vaalyn you don't have to do it that way, easier way would be to add it to `azuracast.env`, then you don't need to specify it in the Docker Compose files. Only reason ports have to be in `.env` is because they're referencing Compose itself, not its containers, but this one is entirely in-container so it's ok to have there.

Vaalyn:
@SlvrEagle23 I've modified the PR and also created another PR for the web container which sets a default to that variable if it is unset.

PR #1635, opened 2019-06-06
Use FFMpeg to analyze media files

Vaalyn:
This is an experimental draft PR that I want to use to investigate if we can improve the performance of the media sync process by using `FFMpeg` instead of `getID3` to analyze the media files.

I've currently only tested this with .mp3 files so I still have to do some testing and probably debugging for other formats.


Vaalyn:
Not worth it

PR #1817, opened 2019-08-05
Add External Public IP as a source for live connection

ErnestPH:
Some users are unable to connect their live broadcasting software when servers are hosted in Cloudflare or other CDN providers. Adding the IP address will help them connect easily.


SlvrEagle23:
@ErnestPH We have a better way of doing this using some internal utilities, because it's not always reliable to check the local or hostname's IP (especially on Docker containers). I will apply that shortly.

SlvrEagle23:
@ErnestPH Implemented via an external-IP-address retrieval service I set up specific to AzuraCast. If you're inside Docker, you basically have to ping out to an external service to pull what the installation's external IP is, so this is all automated now.

PR #1942, opened 2019-09-07
Submit files as array instead of a pipe separated string - Fixes #1932

### LINKED ISSUE #1932: Can't operate with files if there is symbol '|' in filename
Yes

**Host Operating System**
Ubuntu 16.04

**Describe the bug**
Can't add to playlist or delete file if filename has symbol `|`

**To Reproduce**
1. Upload file with filename that contains symbol `|`
2. Try to add to playlist or delete it
3. Nothing happens

Vaalyn:


SlvrEagle23:
@Vaalyn I'm not sure if there was a technical reason it was done this way originally, but it doesn't seem to apply nowadays, so this should work fine. Thank you :D

PR #1999, opened 2019-09-23
replaced admin_username with admin_email

dennybrandes:
In env file its called admin_username but you need to provide an email adress cause frontend is checking it. https://www.azuracast.com/developers/#setting-up-a-local-environment need also to be changed. How to do this?


SlvrEagle23:
@dennybrandes This is more accurate, so I merged it, but just to make sure it's clear, the fixtures and local environment setup are for developing AzuraCast itself, not for using it in production. If you're looking to contribute to the project, by all means go ahead!

PR #2217, opened 2019-12-13
Better quality for mp3

preterive:


SlvrEagle23:
@preterive I've researched this, and the default setting here is 2, which is a solid compromise between MP3 quality and processor load. If we make 0 the default, it will increase the burden of the processor when encoding the MP3 stream, for what is really a very marginal, difficult to observe, improvement in audio quality.

Issue #2474, opened 2020-03-05
Having a space in an advanced url stub breaks it, station boot loop and repeated crashes


SC2Mitch:
**Using Docker installation method**
Yes
**Host Operating System**
Ubuntu Linux 18.04.4
**Describe the bug**
If you leave a space within the Advanced URL Stub, it can cause cascading errors across the station, ranging from the station failing to start, liquidsoap and autoDJ breaking completely. 
**To Reproduce**
Create a station with the advanced URL stub like this
![image](https://user-images.githubusercontent.com/29955942/75939838-5059a380-5e83-11ea-837b-4d9365d5153b.png)

check docker logs and or attempt to start liquidsoap
![image](https://user-images.githubusercontent.com/29955942/75939860-5fd8ec80-5e83-11ea-8748-def17e3940a4.png)

**Expected behavior**
Either, Azuracast to warn users that having a space will cause issues, OR simply prevent the station from being created.
**Relevant Logs**
Station crashing repeatedly
```
stations_1  | 2020-03-05 01:41:01,998 INFO spawned: 'station_76_frontend' with pid 18175
stations_1  | 2020-03-05 01:41:02,049 INFO spawned: 'station_76_backend' with pid 18177
stations_1  | 2020-03-05 01:41:02,051 INFO exited: station_76_frontend (exit status 0; not expected)
stations_1  | 2020-03-05 01:41:02,261 INFO exited: station_76_backend (exit status 1; not expected)
stations_1  | 2020-03-05 01:41:05,113 INFO spawned: 'station_76_frontend' with pid 18181
stations_1  | 2020-03-05 01:41:05,137 INFO exited: station_76_frontend (exit status 0; not expected)
stations_1  | 2020-03-05 01:41:05,162 INFO gave up: station_76_frontend entered FATAL state, too many start retries too quickly
stations_1  | 2020-03-05 01:41:05,495 INFO spawned: 'station_76_backend' with pid 18182
stations_1  | 2020-03-05 01:41:05,683 INFO exited: station_76_backend (exit status 1; not expected)
```
Unable to find a station
```
web_1       | 2020/03/05 01:41:06 [error] 60#60: *14031 open() "/usr/share/nginx/html/pub/test+station" failed (2: No such file or directory), client: 127.0.0.1, server: , request: "POST /pub/test+station HTTP/1.1", host: "localhost:9010"
```
**Screenshots**
https://puu.sh/FhhcY/aefb1be3b6.png

[...]

PR #2477, opened 2020-03-05
Improved warning fix on URL stub

SC2Mitch:
Fixes https://github.com/AzuraCast/AzuraCast/issues/2474


SlvrEagle23:
@SC2Mitch I'd rather actually filter out spaces than just tell people not to use spaces...probably going to take a different approach here.

PR #2513, opened 2020-03-15
Use Shoutcast StreamIDs to specify mountpoint for liquidsoap output

### LINKED ISSUE #2502: Shoutcast2 Multi-Mountpoint doesnt work
Hey there,

I have a test server with Azuracast. There I tried to create a setup with 4 mount points (/stream1 .. 4) on Shoutcast2.

There is the problem that Liquidsoap tries to connect all 4 mount points to the first point (/stream1).

Error in the shoutcast log:
`ERROR [SRC 127.0.0.1:50520 sid = 1] SHOUTcast 1 source rejected. A source is already connected.`

The Azuracast code in Liquidsoap is currently (for /stream2):
`output.icecast (% mp3 (samplerate = 44100, stereo = true, bitrate = 96, id3v2 = true), id = "local_2", host = "127.0.0.1", port = 8000, password = "***** ", mount =" / stream2 ", name =" Name ", description =" Desc ", genre =" Pop ", url =" https://foo-bar.com ", public = false, encoding =" UTF- 8 ", protocol =" icy ", radio)`

So far I have been using Liquidsoap without Azuracast. I used the following working code there:
```
output.icecast (% mp3 (samplerate = 44100, internal_quality = 0, id3v2 = true, bitrate = 96),
id = "stream2",
host = "localhost",
port = 8000,
icy_id = 2,
icy_metadata = "true",
password = "*****: # 2",
name = "name",
url = "https://foo-bar.com",
genre = "Pop", radio)
```

The important difference to Icecast is that the code for Shoutcast2 does not have a `mount` in connection code.

`icy_id` is used for this and the password with the SID Number (Shoutcast Mountpoint) uses:
`hackme:#2`

Maybe you can change it accordingly so that a multi-mountpoint stream is also possible with Shoutcast?

Thank you very much

P.S. Is there someone from the German-speaking world at the developers of Azuracast?

Vaalyn:
This PR resolves #2502 by using the parameters mentioned in the issue.

I've created a test station using the newest free Shoutcast version and created 2 mountpoints to test if the issue is fixed. After applying the changes the station restarts successfully and both mountpoints are playing.

### DIFF
--- a/src/Radio/Backend/Liquidsoap.php
+++ b/src/Radio/Backend/Liquidsoap.php
@@ -1046,7 +1046,7 @@ class Liquidsoap extends AbstractBackend implements EventSubscriberInterface
                 continue;
             }
 
-            $ls_config[] = $this->getOutputString($station, $mount_row, 'local_' . $i);
+            $ls_config[] = $this->getOutputString($station, $mount_row, 'local_', $i);
         }
 
         $event->appendLines($ls_config);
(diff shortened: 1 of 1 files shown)


PB-SF:
Very nice. Tested on my server - all 4 mountpoints will work. Thank you

SlvrEagle23:
@Vaalyn I could swear we already did this at some point in the past, but it must have been either lost to time or part of another side project (like the NowPlaying library itself).

In any case, yep, this is how it's done! Thank you for the PR <3

PR #2550, opened 2020-03-22
Show artist and title on iOS lock screen

### LINKED ISSUE #2547: Wider support for mobile media player display (navigator.mediaSession)
I just discovered a nice feature of the Azuracast player that it updates song metadata for mobile media players. I was testing it on various mobile devices, it seemed to work well on Android with Chrome but no luck on iPhone (tried with Safari and Chrome). 

I found the corresponding code:

https://github.com/AzuraCast/AzuraCast/blob/b4ab44580db7607608db1e948cfe4dd58aaa02b5/frontend/vue/RadioPlayer.vue#L483-L492

and found this (see compatibility): https://developer.mozilla.org/en-US/docs/Web/API/Navigator/mediaSession

Apparently, the feature is not supported by iOS. Is there a way to extend this nice feature for iOS devices too?

Vaalyn:
This PR closes #2547 by binding the `audio` elements `title` attribute to the now playing text.


SlvrEagle23:
@Vaalyn Wow, that actually works? :o

Sweet!

Vaalyn:
@SlvrEagle23 yep, here some screenshots I took when testing this:

DEE829E2-F135-44A3-A3A0-E8E342F2F1E7
9F482BA9-7966-4A5F-A0BC-C1276F691A61

Vaalyn:
@MyTheValentinus Sadly I couldn't find anything in Apples own documentation about this. I've got the solution from this StackOverflow post: https://stackoverflow.com/questions/28303376/is-it-possible-to-change-the-ios-8-lock-screen-audio-label-when-playing-from-web

There is a [webkit bug](https://bugs.webkit.org/show_bug.cgi?id=145411) about implementing the Media Session spec so that Safari can do the same that Android browsers can do with regards to displaying the metadata while playing. This bug has quite a lot of other bugs that it depends on being resolved first but only 3 are currently remaining open. Although the bug is open since 2015 so I wouldn't hope for the feature to land any time soon in webkit.

MyTheValentinus:
Hello @Vaalyn,
Ok, we have same information ^^
Thanks

PR #2836, opened 2020-05-15
Add On-Demand Streaming/Download Support

SlvrEagle23:
This branch represents the initial earliest commit of a new option for radio stations that also want to make a portion of their media library available to the public (such as podcast episodes, royalty-free music, etc.) in an on-demand fashion.

Note: because of royalty restrictions with most radio stations, this feature is turned OFF by default on all stations, and even when enabled, station administrators must select each individual playlist that will contain on-demand available media.


FringeRadioNetwork:
I need the on-demand to load the live stream recordings.  Broadcasting without archives is almost pointless.

SlvrEagle23:
@FringeRadioNetwork This feature relates to offering media in your playlists as on-demand, not the older live broadcasts of DJs. If you would like to see that as a feature, please submit a new feature request on https://features.azuracast.com/

PR #2930, opened 2020-06-05
change OPUS vbr="constrained" to improve sound quality

rj200:
Is there a reason this was set to "none"? That will result in poorer quality sound and is only exists for special use-cases. Constrained mode is the default in libopus and the RFC says it is equivelent to "CBR" mode in MP3 (both maintain a bit reservoir to help with complex moments in the audio.)

References:
https://tools.ietf.org/html/rfc6716#section-2.1.8
https://www.opus-codec.org/docs/opus_api-1.2/group__opus__encoderctls.html#ga34d09ae06cab7e1a6c49876249b67892


SlvrEagle23:
@rj200 The defaults were mostly there out of a lack of understanding of Ogg Opus best practices. I'll take your word for it that `constrained` more accurately represents CBR.

PR #2934, opened 2020-06-06
Change default webDJ bitrate to 320kbps

rj200:
By default webdj encodes audio using libshine at 128kbps, liquidsoap then decodes and re-encodes it in the format of whatever mountpoints you have. libshine is a fast but not high-quality MP3 encoder (it's webpage says it does not have any Psychoacoustic Model!). libshine produces worse quality MP3s at 128kbps than LAME, which is what liquidsoap uses to encode the final stream. I can hear the difference. The result is that webDJ livebroadcasts will sound worse than autoDJ or Butt live broadcasts, even though it look like it's all the same bitrate.

I suggest changing the default to 320kbps to improve quality; this will just increase bandwidth between DJ and azuracast, not from azuracast to listeners. It seems like good practice to keep source audio as high quality a possible until it is finally encoded to send to listeners.

### DIFF
--- a/frontend/vue/webcaster/settings.vue
+++ b/frontend/vue/webcaster/settings.vue
@@ -153,7 +153,7 @@
                 'isStreaming': false,
                 'djUsername': '',
                 'djPassword': '',
-                'bitrate': 320,
+                'bitrate': 256,
                 'samplerate': 44100,
                 'encoder': 'mp3',
                 'asynchronous': true,
--- a/frontend/vue/webcaster/settings.vue
+++ b/frontend/vue/webcaster/settings.vue
@@ -153,7 +153,7 @@
                 'isStreaming': false,
                 'djUsername': '',
                 'djPassword': '',
-                'bitrate': 128,
+                'bitrate': 320,
                 'samplerate': 44100,
                 'encoder': 'mp3',
                 'asynchronous': true,


SlvrEagle23:
@rj200 I understand and agree with your assessment that because Liquidsoap is going to transcode the audio again in any case, sending it a higher-quality audio stream is useful.

However, we do already get a number of complaints from station operators with lower-quality Internet connections that they struggle to maintain the necessary upstream bandwidth to maintain a steady DJ stream, even with high buffering settings. I hesitate to make this situation worse by default, especially since many users might not know to change this.

Maybe...split the difference? 256kbps?

rj200:
Okay. Default 256kbps is probably plenty if it's eventually transcoded down to 128kbps by default.

I actually have an idea I want to try to implement in the WebDJ. Modern web-browsers can natively encode to opus using the MediaRecorder API. Opus is would be practically CD quality at 128kbps. I managed to make a simple prototype streaming "audio/ogg; codec=opus" to azuracast from Firefox. Unfortunately Chrome will only do "audio/webm; codec=opus" and Azuracast's liquidsoap doesn't handle webm container. Perhaps that could be fixed with some liquidsoap config to decode it using  gstreamer of ffmpeg?.

PR #3102, opened 2020-08-22
WIP: Fix issue where scheduled events sometimes played later than they should

cpeter1207:
@SlvrEagle23 this is my first attempt at merging my changes with yours. There is some interaction between the two which causes scheduled events to play way too early now. We need to track that down...


SlvrEagle23:
@cpeter1207 I'm gonna merge this into master and then make a few tweaks (some constants exist elsewhere that can be moved there, making sure the logging handler is popped in all return cases) but otherwise it looks good to me.

PR #3116, opened 2020-08-26
Fixed incorrect decimal, should've been 7,2 from the start

bjarn:
Decimal should've had a scale of 2 from the start. Accidentally changed it to 1 in my previous PR. 

Note to self: do not do PR's late at night anymore :-)


bjarn:
Changed the return type of getLength() to float, otherwise the decimals would never get used in the Liquidsoap config.

PR #3141, opened 2020-09-03
Fix Advanced Play Single Track problem

donbeto97:
Fixed bug for advanced setting PLAY_SINGLE_TRACK not working with scheduled playlists. Missing $this->backendPlaySingleTrack() added.

### DIFF
--- a/src/Entity/StationPlaylist.php
+++ b/src/Entity/StationPlaylist.php
@@ -508,7 +508,7 @@ class StationPlaylist
      */
     public function isPlayable(): bool
     {
-        // Any "advanced" settings are not managed by AzuraCast AutoDJ.  *** FIXED BUG: MISSING $this->backendPlaySingleTrack() ***
+        // Any "advanced" settings are not managed by AzuraCast AutoDJ.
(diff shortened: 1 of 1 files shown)


src/Entity/StationPlaylist.php
```diff
@@ -508,8 +508,8 @@ public function getScheduleItems(): Collection
      */
     public function isPlayable(): bool
     {
-        // Any "advanced" settings are not managed by AzuraCast AutoDJ.
-        if (!$this->is_enabled || $this->backendInterruptOtherSongs() || $this->backendMerge() || $this->backendLoopPlaylistOnce()) {
+        // Any "advanced" settings are not managed by AzuraCast AutoDJ.  *** FIXED BUG: MISSING $this->backendPlaySingleTrack() ***
```
pixel-paul, line 511: It's not appropriate to leave the 'fixed bug' message here, as the commit message explains what you have fixed.

donbeto97: Sorry, Pixel-Paul, I'm new at Github. Thanks for your comments.

PR #3144, opened 2020-09-05
Additional scheduling bug fixes

cpeter1207:


cpeter1207:
@SlvrEagle23 looks like we collided a bit, but I think I resolved the merge conflicts.

I fixed the "looping once-per-day playlist" issue. I also noticed a bug I introduced earlier today where $now would become more and more inaccurate over time. I fixed it by resetting the timestamps in the queue each time a song is dequeued.

I know this is essentially what you were originally doing, but I think my approach is more accurate because it relies on the end of the most recently dequeued song rather than the currently playing song. As I've explained before, there is a race condition where liquidsoap gets the next song in the queue about 30 seconds before it starts playing it. If the buildQueue task runs at that time, it skips an entire song and miscalculates $now.

I'm unsure we even need the buildQueue task anymore, but I address that possibility only resetting timestamps on a dequeue, not every time the buildQueue task runs.

This seems to be performing well for me in my test environment, but I'll keep watching it and submit more refinements as I learn more...

SlvrEagle23:
@cpeter1207 I'm happy to test any improvements out on the scheduler. I think we've reached a state of improved stability relative to the previous iterations, but these little refinements likely will help polish things up better.

PR #3149, opened 2020-09-07
Fix user based locale not working bug

### LINKED ISSUE #3148: Language preference not working?
Yes

**AzuraCast version**
v0.10.3, #de74d50 (2020-09-05 22:00) Rolling • Docker • PHP 7.4

**Host Operating System**
Fresh Minimal CentOS 7 Install

**Describe the bug**
Changing the language to any other than English doesn't change the interface language.

**To Reproduce**
1. Change Language to any other than English

**Expected behavior**
To see the new language.

**Screenshots**
image

**Device(s):**
 - Device: Macbook Pro
 - OS: MacOS 10.15 Catalina
 - Browser:  Chrome 85.0.4183.83 (Official Build) (64-bit)

Tried with 2 different accounts (both Super Administrator) without success.

Vaalyn:
This PR fixes the bug where a user defined locale setting is not working anymore.

The Customization Class is using the `$this->user` property to set the user locale in this line: https://github.com/AzuraCast/AzuraCast/blob/ffa6d6b4fc26499cfc531af34cc9270688bc856e/src/Customization.php#L89-L92

The problem is that `$this->user` is initialized after `initLocale()` is called here: https://github.com/AzuraCast/AzuraCast/blob/ffa6d6b4fc26499cfc531af34cc9270688bc856e/src/Customization.php#L37-L40

This issue was introduced in commit 6f66ff072b2217317dc43f6ddf2c4745f14bcc03

The PR closes #3148

### DIFF
--- a/src/Customization.php
+++ b/src/Customization.php
@@ -34,10 +34,10 @@ class Customization
         $this->settingsRepo = $settingsRepo;
         $this->instanceName = (string)$this->settingsRepo->getSetting(Entity\Settings::INSTANCE_NAME, '');
 
-        $this->locale = $this->initLocale($request);
-
         // Register current user
         $this->user = $request->getAttribute(ServerRequest::ATTR_USER);
+
+        $this->locale = $this->initLocale($request);
 
         // Register current theme
         $queryParams = $request->getQueryParams();


Vaalyn:
@SlvrEagle23 Heya, I've tried to add a new functional test to check if this functionality is working correctly but I'm encountering an issue that prevents that test from working.

This is the test I want to add:
```
/**
 * @before setupComplete
 * @before login
 */
public function changeProfileLocale(FunctionalTester $I)
{
    $I->wantTo('Use a specific locale for a user.');

    $I->amOnPage('/profile/edit');
    $I->see('Edit Profile', '.card-title');

    $I->submitForm('.form', [
        'locale' => 'de_DE.UTF-8',
    ]);

    $I->seeCurrentUrlEquals('/profile');
    $I->see('Deutsch');
    $I->seeInTitle('Mein Account');
}
```

The issue I'm having with that is that when running the tests the `Customization::initLocale()` method will always set the locale for the tester to the `DEFAULT_LOCALE` of `en_US.UTF-8` here: https://github.com/AzuraCast/AzuraCast/blob/ffa6d6b4fc26499cfc531af34cc9270688bc856e/src/Customization.php#L81-L84

I've tried to remove that part and setting the locale of the user that is created by the `@before setupComplete` step like this:
```
...
// Create user account.
$user = new Entity\User;
$user->setName('AzuraCast Test User');
$user->setEmail($this->login_username);
$user->setNewPassword($this->login_password);
$user->getRoles()->add($role);
$user->setLocale('en_US.UTF-8');

$this->em->persist($user);
$this->em->flush();
...
```

But for some reason when I run the tests after doing that the Step `See in source "Logged in"` from the `@before login` starts to fail. (I've added my test to the `A01_Frontend_ProfileCest` so I'm only running that one atm).

What's weird is that the second test does not fail on that step and I'm a bit stumped why that might be since both of them have the `@before setupComplete` and `@before login` annotations.

Here the output from running the tests:

```
- A01_Frontend_ProfileCest: Set a user profile.
✖ A01_Frontend_ProfileCest: Log in to the application. (4.82s)
- A01_Frontend_ProfileCest: Use a specific locale for a user.
✔ A01_Frontend_ProfileCest: Use a specific locale for a user. (1.58s)
```

```
1) A01_Frontend_ProfileCest: Log in to the application.
 Test  tests/functional/A01_Frontend_ProfileCest.php:setProfileInfo
 Step  See in source "Logged in"
 Fail  Failed asserting that  on page /dashboard
--> <!DOCTYPE html>
<html>
<head>
    <meta charset="utf-8">
    <meta http-equiv="X-UA-Compatible" content="IE=edge">
    <meta name="viewport" content="width=device-width, initial-scale=1">

    <title>(Testing) Dashboard - AzuraCast</title>

    <link rel="apple-touch-icon" sizes="180x180" href="/sta
[Content too long to display. See complete response in '/var/azuracast/www/tests/_output/' directory]
--> contains "Logged in".

Scenario Steps:

 4. $I->seeInSource("Logged in") at tests/functional/CestAbstract.php:133
 3. $I->submitForm("#login-form",{"username":"azuracast@azuracast.com","password":"AzuraCastFunctionalTests!"}) at tests/functional/CestAbstract.php:130
 2. $I->seeInCurrentUrl("/login") at tests/functional/CestAbstract.php:126
 1. $I->amOnPage("/") at tests/functional/CestAbstract.php:125
```

Any ideas?

diego-vieira:
@Vaalyn if your test is failing, include the `Functional` test you just did so I can try running on my machine and perhaps see what's going on.

Thanks

diego-vieira:
@Vaalyn there is already a test for that https://github.com/AzuraCast/AzuraCast/blob/master/tests/functional/A01_Frontend_ProfileCest.php

What you can do instead is updating that file with 

```php
<?php
class A01_Frontend_ProfileCest extends CestAbstract
{
    /**
     * @before setupComplete
     * @before login
     */
    public function setProfileInfo(FunctionalTester $I)
    {
        $I->wantTo('Set a user profile.');

        $I->amOnPage('/dashboard');
        $I->see('Dashboard');

        $I->amOnPage('/profile');
        $I->see('Profile');
        $I->see('Super Administrator');

        $I->click('Edit');

        $I->submitForm('.form', [
            'locale' => 'de_DE.UTF-8',
        ]);

        $I->seeCurrentUrlEquals('/profile');
        $I->see('Deutsch');
        $I->see('Sprache');
    }
}
```

Vaalyn:
> @Vaalyn there is already a test for that https://github.com/AzuraCast/AzuraCast/blob/master/tests/functional/A01_Frontend_ProfileCest.php
> 
> What you can do instead is updating that file with
>
>...
> ```

That's the file I mentioned that I had added my test to. But, as I wrote, it does not work because of the `return self::DEFAULT_LOCALE;` in the `Customization::initLocale()` method for test instances. Which is why I'm looking for a solution that might eliminate that part of the method.

I've added a commit to the PR with my test (and changes) included and changed the PR to a draft since it is not yet completely done.

diego-vieira:
@Vaalyn I think that a separate issue for creating this test would be more appropriate than holding the bug fix. I just tested and it's working fine with your changes.

What do you say?

SlvrEagle23:
@diego-vieira The problem with our structure is that if the test suite isn't passing, the Docker image won't build and publish; this is by design to avoid erroneous code making it into the production images.

@Vaalyn If the existing changes you made don't actually let the tests work correctly, you can pull the tests out of this PR and we can merge the fix in directly, then get the fixes back up to spec in a subsequent PR.

Vaalyn:
I've reverted the branch to my first commit that only includes the bugfix and force pushed it.

I'll submit another PR with just the test case as a draft PR.

PR #3160, opened 2020-09-10
Add frontend test case to check if user based locale is working correctly

Vaalyn:
This PR provides a frontend test case that checks if the locale that a user has set for them is working correctly in order to prevent issue #3148 from happening again.

As mentioned in the original PR #3149  for the bugfix this test still has some issues that need to be resolved before we can merge it:

---

For some reason when I run the tests after doing that the Step `See in source "Logged in"` from the `@before login` starts to fail. (I've added my test to the `A01_Frontend_ProfileCest` so I'm only running that one atm).

What's weird is that the second test does not fail on that step and I'm a bit stumped why that might be since both of them have the `@before setupComplete` and `@before login` annotations.

Here the output from running the tests:

```
- A01_Frontend_ProfileCest: Set a user profile.
✖ A01_Frontend_ProfileCest: Log in to the application. (4.82s)
- A01_Frontend_ProfileCest: Use a specific locale for a user.
✔ A01_Frontend_ProfileCest: Use a specific locale for a user. (1.58s)
```

```
1) A01_Frontend_ProfileCest: Log in to the application.
 Test  tests/functional/A01_Frontend_ProfileCest.php:setProfileInfo
 Step  See in source "Logged in"
 Fail  Failed asserting that  on page /dashboard
--> <!DOCTYPE html>
<html>
<head>
    <meta charset="utf-8">
    <meta http-equiv="X-UA-Compatible" content="IE=edge">
    <meta name="viewport" content="width=device-width, initial-scale=1">

    <title>(Testing) Dashboard - AzuraCast</title>

    <link rel="apple-touch-icon" sizes="180x180" href="/sta
[Content too long to display. See complete response in '/var/azuracast/www/tests/_output/' directory]
--> contains "Logged in".

Scenario Steps:

 4. $I->seeInSource("Logged in") at tests/functional/CestAbstract.php:133
 3. $I->submitForm("#login-form",{"username":"azuracast@azuracast.com","password":"AzuraCastFunctionalTests!"}) at tests/functional/CestAbstract.php:130
 2. $I->seeInCurrentUrl("/login") at tests/functional/CestAbstract.php:126
 1. $I->amOnPage("/") at tests/functional/CestAbstract.php:125
```

### DIFF
--- a/src/Customization.php
+++ b/src/Customization.php
@@ -79,9 +79,6 @@ class Customization
     protected function initLocale(?Request $request = null): string
     {
         $settings = Settings::getInstance();
-        if ($settings->isTesting()) {
-            return self::DEFAULT_LOCALE;
-        }
 
         $supported_locales = $settings['locale']['supported'];
         $try_locales = [];
--- a/src/Customization.php
+++ b/src/Customization.php
@@ -65,7 +65,6 @@ class Customization
         $translator->register();
 
         // Register translation superglobal functions
-        putenv('LANG=' . $this->locale);
         setlocale(LC_ALL, $this->locale);
     }
 
@@ -103,6 +102,8 @@ class Customization
         if (!empty($env_locale)) {
             $try_locales[] = substr($env_locale, 0, 5) . '.UTF-8';
         }
+
+        Logger::getInstance()->debug('Locales', ['locales' => $try_locales]);
 
         foreach ($try_locales as $exact_locale) {
             // Prefer exact match.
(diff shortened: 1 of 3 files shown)


SlvrEagle23:
@Vaalyn Figured it out...`Customization` sets the `LANG` environment variable once it determines your locale, and then checks that same environment variable in subsequent checks, which breaks our normal per-request PSR isolation and causes things to persist one request forward from where they would normally persist, which is exactly why this was failing.

Will be working on making sure _all_ pull requests don't just flag as failures in the future, but we can merge this.

PR #3168, opened 2020-09-12
Add possibility to add migrations from plugins via event dispatcher

Vaalyn:
This PR makes it possible to customize the configurations array that is passed to the `Doctrine\Migrations\Configuration\Migration\ConfigurationArray` so that plugins can register their own migrations.

### DIFF
--- a/config/events.php
+++ b/config/events.php
@@ -6,7 +6,7 @@ use App\Middleware;
 use App\Settings;
 
 return function (App\EventDispatcher $dispatcher) {
-    $dispatcher->addListener(Event\BuildConsoleCommands::class, function (Event\BuildConsoleCommands $event) {
+    $dispatcher->addListener(Event\BuildConsoleCommands::class, function (Event\BuildConsoleCommands $event) use ($dispatcher) {
         $console = $event->getConsole();
         $di = $console->getContainer();
 
@@ -31,7 +31,7 @@ return function (App\EventDispatcher $dispatcher) {
             $helper_set->set($doctrine_helpers->get('db'), 'db');
             $helper_set->set($doctrine_helpers->get('em'), 'em');
 
-            $migrateConfig = new Doctrine\Migrations\Configuration\Migration\ConfigurationArray([
+            $migrationConfigurations = [
                 'migrations_paths' => [
                     'App\Entity\Migration' => $settings[Settings::BASE_DIR] . '/src/Entity/Migration',
                 ],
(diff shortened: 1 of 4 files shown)


Vaalyn:
This is an example of how to use this to add the migration path from a plugin in the plugins `events.php`:

```
// Add Migrations to the Doctrine migration_paths
$dispatcher->addListener(Event\BuildMigrationConfigurationArray::class, function (Event\BuildMigrationConfigurationArray $event) {
    $migrationConfigurations = $event->getMigrationConfigurations();
    $baseDir = $event->getBaseDir();

    $migrationConfigurations['migrations_paths']['Plugin\ExamplePlugin\Entity\Migration'] = $baseDir . '/plugins/ExamplePlugin/src/Entity/Migration';

    $event->setMigrationConfigurations($migrationConfigurations);
});
```

Vaalyn:
This is an example of how to add the entities of a plugin to the Doctrine mapping classes paths so that the `bin/console migrations:diff` command can generate the migrations for plugin entities:

```
$dispatcher->addListener(Event\BuildDoctrineMappingPaths::class, function (Event\BuildDoctrineMappingPaths $event) {
    $mappingClassesPaths = $event->getMappingClassesPaths();
    $baseDir = $event->getBaseDir();

    $mappingClassesPaths[] = $baseDir . '/plugins/ExamplePlugin/src/Entity';

    $event->setMappingClassesPaths($mappingClassesPaths);
});
```

One important thing to note here is that the `diff` command currently still generates the migration classes into the `src/Entity/Migration` directory. Not completely sure why that is so atm. one needs to move the generated migrations file from there to their plugin directory.

SlvrEagle23:
@Vaalyn Does `array_unshift`ing the value onto the front of the array instead fix it?

Vaalyn:
> @Vaalyn Does `array_unshift`ing the value onto the front of the array instead fix it?

Changing the order of the `$mappingClassesPaths` entries does not have any effect. When I change the order of the `migrations_paths` however all migrations that `migrations:diff` generates are put into the `Migration` directory of the plugin instead of the one of the application. Regardless of which entity I modify.

If I'm understanding it correctly the Doctrine Migrations 3.0 release has made it possible to add additional `migrations_paths` but they haven't updated the `diff` command to automatically generate migrations into the corresponding paths yet. See the following comments in [the PR](https://github.com/doctrine/migrations/pull/858) for v3.0:

> Multiple namespaces are allowed (and paths). This allows to provide third party migrations, plugin-provided migrations and and dependent migrations as well ...

and further down in the PR:

> Rollup and Diff commands have not been updated yet.

I just noticed however that there is a `namespace` option on the `diff` command and when I explicitly use the namespace of my plugins migrations it correctly generates them into the plugins `Migration` directory.

That's how it works for me:
```
bin/console migrations:diff --namespace "Plugin\ExamplePlugin\Entity\Migration"
```

Vaalyn:
After playing around a bit more I think this PR should be ready to be merged.

The only thing that someone needs to actively do to correctly generate the migrations into the correct namespace is using the `--namespace` parameter of the `migrations:diff` command like I mentioned. When we add an example to the readme of the Example Plugin this needs to be included there.

In order for the plugins entities to correctly work one also needs to generate the proxy classes for the entities and clear the cache(s). I have not yet gone through a "normal" plugin installation process since I'm still working on the plugin I'm testing this with so I'm not sure if that will do all those things but I will test that and if there is something missing I'll add that in another PR.

PR #3172, opened 2020-09-13
Add configurable time range for duplicate prevention

Vaalyn:
This PR implements a user configurable time range (in minutes) that the duplicate prevention algorithm should take into account.


Vaalyn:
I've also found that the calculation of the "least played song" that is done when a duplicate can't be prevented is not really accurate due to a small bug. Since this part of the `preventDuplicates()` method only takes uses the timestamp of the results of the `station_queue` it does not actually get the "least played" song: https://github.com/AzuraCast/AzuraCast/blob/master/src/Radio/AutoDJ/Queue.php#L428

PR #3231, opened 2020-10-04
Remove the `Song` entity and restructure dependent tables accordingly

SlvrEagle23:
AzuraCast's predecessor, Ponyville Live, had a single parent `song` table that contained an entry for every unique song ever played across all stations. The idea behind this was so that if a station wanted us to change how the song wound up displaying on our players, we could edit the entry in the `songs` table and it would update automatically every time a song with that ID was played anywhere in the system.

With today's AzuraCast, this table is entirely vestigial for a number of reasons, the most important being that the `songs` table is not itself editable, can expand to incredibly large sizes for no added benefit, and a track that winds up having the same "hash" may actually be intended to display differently between two different stations.

This PR removes the `songs` table entirely and denormalizes its contents out to the tables that depended on it, `station_queue`, `station_media` and `song_history`. All three entities now extend from a base `Song` class that carries the same information that the previous separate `Song` table carried, allowing for a smooth transition of sections of code that expected that content to be there, but removing an expensive join from many queries across the system.


Vaalyn (commented):
Like the addition of configurable log levels. Could help us getting more information of errors that we might not be able to reproduce. 👍 

------

The use of an instantiable parent class `Song` is a bit strange for me since the only reasons for having it instantiable that I see is the `NowPlaying` sync task and for sending a `new Entity\Song('Remote Playlist URL')` to the constructor of the `StationQueue` in the `Radio\AutoDj\Queue` in the `playSongFromPlaylist()` method.

In the case of the `NowPlaying` sync task there are 2 places where the `Song` entity is used.

1. To be able to call `getSongApi()` to get an `Api\Song` object back. This object will only have `text` set to the `Stream Offline` string (if I'm not mistaken here) and `art` as well as `custom_fields` but those are coming from the provided `$api_utils`.

2. To call `$this->history_repo->register(new Entity\Song($npResult->currentSong), $station, $np);`.

I find it weird to have a Doctrine Entity class which is not persistable but is instantiable and that is (basically) only used to instantiate data for child classes from a broad set of input data or to use it to get access to stuff that methods of that class return. It seems like the entity / base class has responsibilities that it shouldn't have which is (in my opinion) further hinted at by the logic in the setter `setSong` and it's rather permissive / wide parameter type of `self|Api\Song|CurrentSong|array|string`.

I would prefer having an interface that is implemented by the entities that (currently) extend `Song` and a trait to add those shared fields / methods to those entities. That's a concept that I have seen in multiple big projects but I've never encountered the constellation used here.

Since this is (in my opinion) introducing some technical debt here it might be worth talking about this. 

(If this intended to be an intermediate state which will be overhauled again later, could we create an issue for that?)

SlvrEagle23:
@Vaalyn I realize I already merged this but I'm happy to discuss its implications here in any event since there's really no part of AzuraCast that I consider "gospel" or set in stone.

Also, the addition of the configurable log levels here instead of being a separate PR is a mistake on my part that I didn't want to bother rebasing. 😆 

The reason I made `Song` a base class instead of using traits/interfaces is because sometimes it's useful to actually pass around a `Song` object by itself, and the application does so in a number of spots without ever persisting them to the database; this allows for that to continue to happen without needing to create one of the more specific entity classes.

I can see how it looks like it's introducing unnecessary complexity and/or it's something we don't already do, but we also have non-persisted entities like `StationBackendConfiguration` and `StationFrontendConfiguration` in that same folder already, since the `Entity` folder is really just meant to hold any models, regardless of whether they persist to the database or not.

SlvrEagle23:
@Vaalyn I took another look at all of this and hopefully in the latest commit I improved the "code smell" quite a lot by implementing a `SongInterface` and `HasSongFields` trait and modifying the constructor/`setSong` function to only accept a `SongInterface` implementation (with some other little converters built in as static functions on the `Song` entity itself, which is intended solely to be transiently used to compare against known hashes.

Vaalyn:
> Also, the addition of the configurable log levels here instead of being a separate PR is a mistake on my part that I didn't want to bother rebasing. 😆

You are forgiven 😛 

> The reason I made Song a base class instead of using traits/interfaces is because sometimes it's useful to actually pass around a Song object by itself, and the application does so in a number of spots without ever persisting them to the database; this allows for that to continue to happen without needing to create one of the more specific entity classes.

It is indeed useful to be able to pass around a `Song` object for the parts of the application where we do this and I don't refute this. I also don't want to argue for "always create an instance of a specific entity" instead. For me the combination of having an entity that's not persitable, using it for passing around shared data and defining columns for child classes is a violation of the Single Responsibility Principle.

> I took another look at all of this and hopefully in the latest commit I improved the "code smell" quite a lot by implementing a SongInterface and HasSongFields trait and modifying the constructor/setSong function to only accept a SongInterface implementation (with some other little converters built in as static functions on the Song entity itself, which is intended solely to be transiently used to compare against known hashes.

I've looked at the refactoring and I like it. That looks much cleaner to me.

> I can see how it looks like it's introducing unnecessary complexity and/or it's something we don't already do, but we also have non-persisted entities like StationBackendConfiguration and StationFrontendConfiguration in that same folder already, since the Entity folder is really just meant to hold any models, regardless of whether they persist to the database or not.

In the case of `StationBackendConfiguration` and `StationFrontendConfiguration` I view it a bit differently since those are persisted as JSON strings as part of the `Station` entity (we don't use something like doctrine-json-odm but it's similiar enough for me to count to the same category).

The `Song` object is more of a DTO than an entity. I know we have no dedicated DTO namespace since we don't really have any other objects that classify as DTOs in the application and the `StationBackendConfiguration` and `StationFrontendConfiguration` are more of a `JsonDocument` type than an entity type.

> the Entity folder is really just meant to hold any models, regardless of whether they persist to the database or not

I understand that and it's not a big architectural issue for me. I just think new developers could stumble upon this because in most cases entities refer to database entities. Having a clear separation between database entities and those other objects could improve the clarity of the code further. When I see a part of the code with `new Entity\Something()` the first thing I'd assume is that this is something that is persisted to the database in it's own table since it's from the root `Entity` namespace. Apart from those 3 classes we don't have any other models in the `Entity` namespace and the "new" `Song` class is, well, new so I think this could be an ideal opportunity to improve this a little bit.

----

One small thing that I wanted to ask about the refactoring is the change of the method name `getSongApi` to `api`. I'm a big fan of self describing / meaningful naming and a method with the name `api` does not really tell the reader what this method is doing. I think `toApiSong` would be a better name for that method since it's transforming the instance of this object to an instance of `Api\Song`.

----

(I hope I'm not coming of as pedantic regarding these things. Since you've requested a review from me I'm trying to give this as much attention as I can and I've learned to always point out things that might be worth talking about even when they are relatively small.)

SlvrEagle23:
@Vaalyn The move from `getSongApi` to `api` was actually just reverting _back_ to the standard across all models where `->api()` produces an API-friendly representation of that model. We could rename it, sure, but it's consistent across the system and was only disrupted because of the class inheritance on `Song` messing it up, which became a non-issue with the interface/trait switch.

Vaalyn:
> @Vaalyn The move from `getSongApi` to `api` was actually just reverting _back_ to the standard across all models where `->api()` produces an API-friendly representation of that model. We could rename it, sure, but it's consistent across the system and was only disrupted because of the class inheritance on `Song` messing it up, which became a non-issue with the interface/trait switch.

👍 Thank you for the clarification.

The point of improving the self describing nature of the method name still stands but is then definitely not something that should be discussed here.

PR #3239, opened 2020-10-05
Switch from SweetAlert to SweetAlert2

limonte:
I would like to propose to switch from SweetAlert to [SweetAlert2](https://github.com/sweetalert2/sweetalert2) - the supported fork of SweetAlert. The original reason for creating SweetAlert2 is inactivity of SweetAlert: https://stackoverflow.com/a/27842854/1331425 

### Reasons to switch:

1. Accessibility (WAI-ARIA) - SweetAlert2 is fully WAI-ARIA compatible and supports all popular screen-readers. Accessibility is a must in 2020, there are a lot of explanatory and tech articles, but this one is truly inspirable: [Software development 450 words per minute](https://www.vincit.fi/en/blog/software-development-450-words-per-minute/)

2. Better buttons contrast is a huge advantage for all users especially for users with vision disabilities. 

3. Better support, average time to resolve an issue:

   - SweetAlert: 
   - SweetAlert2: 

4. SweetAlert2 is far more popular and battle-tested that SweetAlert:

   - SweetAlert: 
   - SweetAlert2:

### DIFF
--- a/config/assets.php
+++ b/config/assets.php
@@ -159,7 +159,7 @@ return [
         'files' => [
             'js' => [
                 [
-                    'src' => 'dist/lib/sweetalert/sweetalert.min.js',
+                    'src' => 'dist/lib/sweetalert2/sweetalert2.all.min.js',
                     'defer' => true,
                 ],
                 [
--- a/frontend/gulpfile.js
+++ b/frontend/gulpfile.js
@@ -48,10 +48,10 @@ var jsFiles = {
       'node_modules/bootstrap-notify/bootstrap-notify.min.js'
     ]
   },
-  'sweetalert': {
-    base: 'node_modules/sweetalert/dist',
+  'sweetalert2': {
+    base: 'node_modules/sweetalert2/dist',
     files: [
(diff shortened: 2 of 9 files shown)


SlvrEagle23:
@limonte Hello and thank you for your contribution, as well as your ongoing maintenance of the SweetAlert2 library!

I merged your PR only to notice after the fact that SweetAlert2 has a slightly different return structure, and rather than returning a simple boolean value, it returns an object with the `.value` property as a boolean instead. This was causing all of our buttons across the application to trigger regardless of whether the user clicked "Cancel" or the confirmation button, since we were checking for the base return value, which always exists now.

I've pushed a fix to our master branch that resolves the issue, but if you submit any PRs to other projects that retrofits their installations to use SweetAlert2 this may be something you want to include in the original request.

limonte:
Hey @SlvrEagle23 and thank you for the prompt review and merge!

Apologies for missing the `.value` part, my mistake. As an alternative you can also use `.isConfirmed` which is documented here: https://sweetalert2.github.io/#handling-buttons

PR #3243, opened 2020-10-07
Analytics Overhaul and InfluxDB Removal

SlvrEagle23:
Summary of changes made in this PR:

 - The `Analytics` table and entity has been expanded:
    - The `timestamp` column is now a `DateTime` that Doctrine automatically converts into a `CarbonImmutable`
    - The `number_avg` column is now a `decimal` field, allowing proper storage of averages
    - A new `number_unique` column has been created to track unique listeners per hour/day
    - New constants for the measurement intervals (hour/day)
 - The `Sync/Tasks/Analytics` task has been rewritten to directly ingest analytical data into the `Analytics` entity based on the contents of the `SongHistory` and `Listener` entities
 - InfluxDB is removed from the entire application dependency stack, and its PHP library is removed from the application
 - The dashboard statistics page and the "Statistics Overview" page (the only two pages that use InfluxDB's data) have been rewritten to use the `Analytics` database
 - The `Listener` table is now cleaned up at the same frequency as `SongHistory` which is an oversight we've missed that has likely caused extreme ballooning of some databases
 - The "waiting for services to start up" Dockerize script has been replaced with a much friendlier pure-PHP one in `bin/uptime_wait` that has a cute little radio animation


bin/uptime_wait
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    protected function println(string $line): void
+    {
+        echo $line . "\n";
+    }
+}
+
+$uptimeWait = new UptimeWait;
+$uptimeWait->run();
```
Vaalyn, line 139: General question because I've noticed that there is no newline at the end of some files in AzuraCast and also in the new `SongInterface`.

Have you configured your IDE to automatically add a newline at the end of files when saving the file?

Since the overwhelming majority of files in AzuraCast have a newline at the end I'd advise you to enable that option.

src/Controller/Frontend/DashboardController.php
```diff
@@ -134,7 +130,7 @@ public function indexAction(ServerRequest $request, Response $response): Respons
 
             $cache_name = 'homepage_metrics_' . implode(',', $stats_cache_stations);
 
-            $metrics = $this->cache->get($cache_name);
+            // $metrics = $this->cache->get($cache_name);
             if (empty($metrics)) {
                 $metrics = $this->_getMetrics($view_stations, $show_admin);
                 $this->cache->set($cache_name, $metrics, 600);
```
Vaalyn, line 136: Is the loading of the metrics from the cache intentionally commented out?
I can see that after they are loaded they are send to the cache so I assume that the `//` is not intended to be there.

bin/uptime_wait
```diff
@@ -0,0 +1,139 @@
...
+            '🖥️🎶－🎵－📻',
+            '🖥️－🎶－🎵📻',
+            '🖥️🎵－🎶－📻',
+            '🖥️－🎵－🎶📻',
+        ]);
+
+        $_ENV = getenv();
+
+        $applicationEnv = $_ENV['APPLICATION_ENV'] ?? 'production';
+        $this->debugMode = ('production' !== $applicationEnv);
```
Vaalyn, line 63: Could make sense to base this on the `LOG_LEVEL` environment variable so that users can selectively get more debug information without changing their app environment.

bin/uptime_wait
SlvrEagle23: @Vaalyn This was turned off in `.editorconfig` so I just turned it back on, should fix it up.

src/Controller/Stations/Reports/OverviewController.php
```diff
@@ -45,41 +44,41 @@ public function __invoke(ServerRequest $request, Response $response): ResponseIn
...
         $daily_chart = new stdClass;
         $daily_chart->label = __('Listeners by Day');
         $daily_chart->type = 'line';
         $daily_chart->fill = false;
 
         $daily_alt = [
-            '<p>' . $daily_chart->label . '</p>',
-            '<dl>',
+            ' <p>' . $daily_chart->label . ' </p> ',
+            '<dl> ',
```
Vaalyn, line 60: What purpose do these additional spaces serve here?

src/Entity/Analytics.php
```diff
@@ -34,15 +40,24 @@ class Analytics
...
      */
     protected $type;
 
     /**
-     * @ORM\Column(name="timestamp", type="integer")
-     * @var int
+     * @ORM\Column(name="timestamp", type="carbon_immutable", precision=0)
+     * @var CarbonImmutable
      */
     protected $timestamp;
```
Vaalyn, line 62: Since this is no longer a simple timestamp value I think we should change the name of this field. Leaving it as `timestamp` could potentially be confusing since it is a `DateTime` in the DB now. I'd propose something like  `point_in_time`, `moment`, `entry_datetime`,.

src/Entity/Analytics.php
```diff
@@ -59,34 +74,41 @@ class Analytics
     protected $number_max;
 
     /**
-     * @ORM\Column(name="number_avg", type="integer")
-     * @var int
+     * @ORM\Column(name="number_avg", type="decimal", precision=10, scale=2)
+     * @var float
```
Vaalyn, line 78: The doctrine type `decimal` is mapped to a `string` in PHP according to the Doctrine documentation:
> decimal: Type that maps a SQL DECIMAL to a PHP string.

src/Entity/Analytics.php
```diff
@@ -114,13 +143,13 @@ public function getNumberMax(): int
         return $this->number_max;
     }
 
-    public function getNumberAvg(): int
+    public function getNumberAvg(): float
```
Vaalyn, line 146: See my comment about the doctrine to PHP type mapping above. Returning the `number_avg` as is should fail here. This should either return a string or cast the variable to float.

src/Entity/Fixture/Analytics.php
```diff
@@ -0,0 +1,72 @@
...
+        $em->flush();
+    }
+
+    public function getDependencies()
+    {
+        return [
+            Station::class,
+        ];
+    }
+}
```
Vaalyn, line 72: Missing newline at end of file

src/Entity/Repository/AnalyticsRepository.php
```diff
@@ -0,0 +1,26 @@
...
+            WHERE a.station = :station
+            AND a.type = :type
+            AND a.timestamp >= :threshold')
+            ->setParameter('station', $station)
+            ->setParameter('type', $type)
+            ->setParameter('threshold', $threshold)
+            ->getArrayResult();
+    }
+
+}
```
Vaalyn, line 26: Missing newline at end of file

src/Middleware/Module/Stations.php
```diff
@@ -29,8 +29,6 @@ public function __invoke(ServerRequest $request, RequestHandlerInterface $handle
         $backend = $request->getStationBackend();
         $frontend = $request->getStationFrontend();
 
-        date_default_timezone_set($station->getTimezone());
```
Vaalyn, line 32: Small question so I can better understand it.

What part of the code depends / depended on the value of `date_default_timezone_set()` before this PR and after the PR? Since it's now set in the `AppFactory` to always be `UTC` I'd like to know more about the dependencies of that value.

src/Service/UptimeWait.php
```diff
@@ -4,7 +4,6 @@
 use Doctrine\DBAL\Connection;
 use Doctrine\ORM\EntityManagerInterface;
 use Exception;
-use InfluxDB;
 use Redis;
 
 class UptimeWait
```
Vaalyn, line 9: Is this class used anywhere after this PR? I've just done a quick search over `master` for `UptimeWait` and saw that the only places this class was used previously are the `SyncCommand` and `UptimeWaitCommand` which are deleted in this PR.

src/Middleware/Module/Stations.php
SlvrEagle23: @Vaalyn I did some research and it turns out that if you do `CarbonImmutable::now()` with no timezone specified, it uses the timezone that's currently the default via this function. While we try to expressly avoid doing this and always try to either load a timezone in the station's time or UTC, having the default change from UTC anywhere in the application certainly invites some hard-to-diagnose problems for common code that may be initializing a new DateTime/Carbon instance without a timezone specified.

Vaalyn: Thank you for the explanation, nice catch btw.

src/Sync/Task/Analytics.php
```diff
@@ -2,123 +2,231 @@
...
+                break;
+
+            case Entity\Analytics::LEVEL_ALL:
+                $this->updateAnalytics(true);
+                break;
         }
     }
 
-    protected function purgeAnalytics(): void
+    protected function updateAnalytics(bool $withListeners = true): void
```
Vaalyn, line 30: There is quite a lot going on in this method. It took me longer to step through everything in my head than I'd like to admit.

I think the this method would benefit from splitting it into multiple `protected` methods to separate some of the code.

Another thing that would help reduce the amount of code here would be to move the query building into the `AnalyticsRepository` by providing methods that return the values needed here / do the database actions like `deleteAnalyticsForLastDays(int $numberOfDays): void`, `createHourlyAnalyticsEntry(...): Analytics` and `createDailyAnalyticsEntry(...): Analytics`. So that the Sync Task itself is only really doing the calculations and hands off all the other stuff.

src/Controller/Stations/Reports/OverviewController.php
```diff
@@ -45,41 +44,41 @@ public function __invoke(ServerRequest $request, Response $response): ResponseIn
...
         $daily_chart = new stdClass;
         $daily_chart->label = __('Listeners by Day');
         $daily_chart->type = 'line';
         $daily_chart->fill = false;
 
         $daily_alt = [
-            '<p>' . $daily_chart->label . '</p>',
-            '<dl>',
+            '<p>' . $daily_chart->label . ' </p>',
+            '<dl> ',
```
Vaalyn, line 60: Still 2 spaces here that you might have overlooked. If they are intentional please let me know.

src/Sync/Task/Analytics.php
SlvrEagle23: @Vaalyn I agree to some extent, though if we move some of the prepared queries from the task to repositories, we don't get to take advantage of the fact that the query's DQL is only parsed once by Doctrine instead of needing to be parsed multiple times. (I think. I'm not sure that there's any actual performance gain there.)

src/Controller/Stations/Reports/OverviewController.php
SlvrEagle23: @Vaalyn Pretty sure this is my IDE being overly clever after pasting some HTML in from another source. Will fix.

src/Entity/Analytics.php
Vaalyn: I think this might have been accidentally marked as resolved by one of the last commits.

src/Entity/Analytics.php
Vaalyn: I think this might have been accidentally marked as resolved by one of the last commits.

SlvrEagle23: @Vaalyn Nope, notice under it I actually cast the string to a float before returning now.

Vaalyn: Yeah, GitHub's interface trolled me...

src/Sync/Task/Analytics.php
Vaalyn: I don't think that they will be parsed multiple times since the queries itself wouldn't change when calling the Repository methods. The only thing that changes between the calls are the parameters and those are not parsed into the final Query so the parsed DQL should still be in the Doctrine Query Cache. Anyways I'd prioritise clarity first and if we notice performance issues here then we should try to optimize that.

SlvrEagle23: @Vaalyn Latest commit moves a lot of this out to the repositories (there actually already was a "Get Unique listeners in this time period" repo function for Listeners, so it deduplicated some code there), hopefully helping the cognitive load of that section.

Vaalyn:
What exactly is being done with the installed `InfluxDB` service on Ansible installations when it's removed from the Ansible tasks? Is it left running, is it just stopped or automatically being uninstalled?
Since for our Docker users the service wouldn't exist after an update that should be the desired state for our Ansible users as well.

SlvrEagle23:
> What exactly is being done with the installed `InfluxDB` service on Ansible installations when it's removed from the Ansible tasks? Is it left running, is it just stopped or automatically being uninstalled?
> Since for our Docker users the service wouldn't exist after an update that should be the desired state for our Ansible users as well.

Latest commit actively removes it from updating installations; it's still removed from new installations as it simply isn't needed, but both updates and uninstalls will still remove it if it's there.

src/Sync/Task/Analytics.php
```diff
@@ -2,123 +2,215 @@
...
+                        $unique = $this->listenerRepo->getUniqueListeners($station, $start, $end);
 
-        foreach ($results as $stat_series => $stat_rows) {
-            $series_split = explode('.', $stat_series);
-            $station_id = ($series_split[1] === 'all') ? null : $series_split[1];
+                        $hourlyUniqueListeners ??= 0;
+                        $hourlyUniqueListeners += $unique;
+                    } else {
+                        $unique = null;
+                    }
```
Vaalyn, line 97: This `else` is unnecessary if you put `$unique = null;` before the `if ($withListeners) {`.

src/Entity/Analytics.php
```diff
@@ -59,34 +74,43 @@ class Analytics
...
-     * @var Station|null
+     * @ORM\Column(name="number_unique", type="integer", nullable=true)
+     * @var int|null
      */
-    protected $station;
+    protected $number_unique;
 
     public function __construct(
+        DateTimeInterface $moment,
         Station $station = null,
```
Vaalyn, line 90: Could we make the parameter type declaration for `$station` nullable here? I know that it doesn't make a difference for PHP (regardless of strictness) but since we actually want to supply it with `null` values sometimes the constructor should indicate that directly. The `$number_unique` parameter for example is already declared to be nullable in this constructor.

src/Sync/Task/Analytics.php
```diff
@@ -2,123 +2,213 @@
...
-                    $row['station_id'] ? $stations_by_id[$row['station_id']] : null,
-                    $row['type'],
-                    $row['timestamp'],
-                    $row['number_min'],
-                    $row['number_max'],
-                    $row['number_avg']
+                $dailyMin = min($dailyMin, $dailyStationMin);
+                $dailyMax = max($dailyMax, $dailyStationMax);
+
+                $dailyStationUnique = null;
```
Vaalyn, line 159: Would you mind adding a `Listeners` at the end of the variable to clarify that this holds the listeners?
For reference look at the `$dailyUniqueListeners` variable.

Vaalyn (approved):
👍 LGTM

PR #3266, opened 2020-10-11
Add "genre" as a first-class StationMedia meta field

### LINKED ISSUE #3214: Make "Genre" a First-Class Station Media Metadata Item
Currently, it is possible to implement "Genre" via the Custom Fields functionality, and to have it automatically set via an ID3 variable, but this doesn't treat it as a "first-class" field the same way that natively adding it to the `StationMedia` entity would.

This feature request is to:
 - Add `genre` as a property in `src/Entity/StationMedia.php`
 - Create the corresponding getters/setters for the new property
 - Create a new database migration adding the "genre" field
 - Update the `frontend/vue/StationMedia.vue` editor panel to include the field, and to have it as a selectable column in the overall table
 - Update the API responses (populated in the `StationMedia` entity) to include `genre`.

xewl:
I've set 30 as the length, given that's the length of ID3v1 Enhanced TAG for genre.

Ok, let's fix that test's error too. :') 

closes #3214


SlvrEagle23:
@xewl I realize I didn't explain this in the original issue, but I actually _don't_ want this to be a part of `SongInterface`, which is only meant to contain the most basic song playback information that's always transmitted with Icecast/Shoutcast (text/artist/title), not the richer metadata that we attach to the media.

This should probably specifically be a property on `StationMedia` and the `Entity\Api\Song` response only, not the `SongInterface` and the `HasSongFields` trait.

xewl:
@SlvrEagle23 I noticed my mistake. No worries :)
I'm setting up Docker/WSL2 rn (I was still rocking v1) so I can run the tests locally 😬

SlvrEagle23:
@xewl No worries, will switch this over to draft status while it gets polished up. Thanks for working on it!

src/Entity/StationMedia.php
```diff
@@ -236,6 +245,16 @@ public function setAlbum(?string $album = null): void
...
     }
 
+    public function getGenre(): ?string
+    {
+        return $this->genre;
+    }
+
+    public function setGenre(?string $genre = null): void
+    {
+        $this->genre = $genre;
```
SlvrEagle23, line 255: @xewl Make sure to call `$this->truncateString($genre, 30)` to avoid the MariaDB "text truncated for column" error.

xewl: Good call. Will do.

xewl:
Off-topic: I have some git stuff to learn, dang 🤣

src/Entity/Migration/Version20201010170333.php
```diff
@@ -0,0 +1,27 @@
...
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20201010170333 extends AbstractMigration
+{
+    /**
```
SlvrEagle23, line 12: @xewl This migration should've generated with strict types and a description field. Are you on the latest version of the Doctrine Migrations library that we're using?

xewl: I might've created this manually. 😬 
Will look into it

xewl: Honestly, the dev setup isn't working for me rn. Docker/WSL2 troubles I can't seem to get around.
I've edited the migration to resemble the last 2. Hope that's fine..

xewl:
This can stay in draft for a day or 2, that is: if anything else needs to be changed.
> I can't seem to get around WSL2/Docker configuration troubles rn.

SlvrEagle23:
@xewl Everything looks good; only thing I'd want to add would be to automatically set the genre via ID3 tags in `App\Entity\StationMediaRepository::loadFromFile` and to write it to the file via the `writeToFile` in the same repo.

xewl:
I'm currently assuming it's going to return a string... I'd have to look deeper.

SlvrEagle23:
@xewl Ah, didn't know about the whole "genreid" thing...we may be able to modify our writing into two separate processes, one for id3v2 and another (with genre removed) for id3v1.

xewl:
@SlvrEagle23 Or ignore v1's arbitrary thing altogether? :3 They didn't seem to like the approach themselves. xD

SlvrEagle23:
@xewl Also fine by me. We tend to heavily prefer id3v2 anyway.

xewl:
Looks like for v1 it gets parsed into an ID anyway, or defaults to `255`
https://github.com/JamesHeinrich/getID3/blob/79c467acde0db96edcd3ab9c0b8d4b68039f8f20/getid3/write.php#L550

I think we're pretty clear then.

xewl:
Alright, I've "fixed" my docker&wsl2 (some IP Loopback issue still)
But.. I was finally able to set up the dev environment to test my previous and ongoing changes..

Noticed I still have some work @ the Media Edit form.
The parsing/reading of tags did work though :D

To be continued in the next 2 days when I find the time :)

xewl:
@SlvrEagle23 The last test that failed was a phpcs rule because the line length now exceeds 122 chars.
As I can't find any examples of the same case in any other of the Azura repositories, I'm wondering if you're fine with this:

```php
            ->select('partial sm.{
                id,
                unique_id,
                art_updated_at,
                path,
                length,
                length_text,
                artist,
                title,
                album,
                genre
            }')
```

PR #3341, opened 2020-10-27
Unified Filesystem Overhaul

SlvrEagle23:
The goal of this project and pull request is to accomplish several major changes in AzuraCast that have been a long time coming and will make the experience of the system significantly better, especially for stations sharing their libraries:

 - To have all media associate with a "Storage Location" instead of an individual station; this means that if media is shared between multiple stations, it's only processed _once_ and is available to all of those stations immediately, and the file space occupied by album art and waveforms is no longer duplicated between stations.

 - The newly-created "Storage Location" entities can be local or remote, with initial support planned for Amazon S3 API supported platforms (including DigitalOcean Spaces).

 - These new "Storage Spaces" will also be implemented for two other high-capacity items: backups and recordings of live broadcasts.

 - To support the technical underpinnings of this change, switching to League/Flysystem 2.0.

As this is a fundamental change to how AzuraCast is structured, the plan is for this draft to be open for quite a while for rigorous testing and review.

Major things that will need testing will include:

 - The new UI/UX for managing storage spaces, and the modifications to the station form to allow administrators to select the appropriate storage group for each station.

 - The performance and functionality of remote S3 connections, especially how the functionality of copying briefly to a temporary local filesystem works with Liquidsoap.


src/Entity/Migration/Version20201027130404.php
```diff
@@ -0,0 +1,160 @@
...
+
+        $this->addSql('ALTER TABLE station_media DROP FOREIGN KEY FK_32AADE3A21BDB235');
+        $this->addSql('DROP INDEX IDX_32AADE3A21BDB235 ON station_media');
+        $this->addSql('DROP INDEX path_unique_idx ON station_media');
+        $this->addSql('ALTER TABLE station_media ADD storage_location_id INT NOT NULL');
+    }
+
+    public function postUp(Schema $schema): void
+    {
+// Deleting duplicate streamers to avoid constraint errors in subsequent update
```
Vaalyn, line 37: small copy/paste relic comment it seems 😉

SC2Mitch:
I'll be happy to help with debugging and testing if you need a extra pair of eyes.

Vaalyn:
Looking good so far 👍

src/Entity/Station.php
```diff
@@ -657,81 +629,67 @@ public function setRadioBaseDir(?string $newDir = null): void
         $this->radio_base_dir = $newDir;
     }
 
-    public function getRadioAlbumArtDir(): string
+    public function checkDirectories(): void
```
Vaalyn, line 632: The name of this method seems a bit vague to me since it's not only checking directories but rather creating them if they don't exist. I'd like to propose renaming it to `ensureDirectoriesExist` to make it clear that this is what the method does.

SlvrEagle23:
I think we're at a reviewable state at this point, everything is green-lit.

Ready for testing @Vaalyn and @SC2Mitch!

SlvrEagle23:
Never mind, missed the last little bit of ensuring Liquidsoap copies S3 bucket locations to local when playing.

Needs a little more work.

SlvrEagle23:
The remaining portion (handling the transfer from S3 to a temporary directory for Liquidsoap and optimizing how S3 works across the system) is complete. PR is ready for review and testing again.

Vaalyn:
Have already looked at the source code changes but didn't have the time yet to thoroughly test and play around with it.

Will do so over the weekend. Just wanted to add this comment so that you know I'm still on it.

Vaalyn (changes requested):
I've been playing around with the S3 bucket that you provided me with @SlvrEagle23 and in general it's working quite well.

There are some things that I've noticed that I'd like to point out though:

- When creating an S3 StorageLocation it should be tested to check if accessing it with the provided credentials is working
	- I've accidentally created it wrongly which lead to me seeing a 500 error on the `http://localhost/api/station/7/files/list` request (in the browser console, not the page itself) when looking at the music files for the station. Doesn't make it easy to see that there is a config problem with the storage location.

- "Credential Key" and "Credential Secret" should be renamed to "Access Key ID" and "Secret Key" for better separated terminology

- Improve Location dropdowns to be better readable
	- The path + "(Station xx #x)" looks convoluted, is the `(Station xx #x)` part necessary?
	- Local path should begin with `/var` instead of `var`
	- Suggestion: https://prnt.sc/vfape2 and https://prnt.sc/vfar6p

- Rename "Amazon S3" to just "S3" because it makes it look like it's specifically for Amazon S3 instead of the general S3 protocol (I know it's mostly from Amazon but I see the potential to cause confusion and lead people to think it does not support other S3 providers)

- `cleanStationTempFiles` cleans up files older than 2 days meaning that when we have let's say ~2mb per 1 minute of audio we have around 2.8GB per day so 5.6GB of temporary data that is potentially always used locally by the files pulled from S3. (per station)
	- I think we should make this timerange configurable.
		- (The following point is only applicable when the local copies are re-used when a song is played again which is currently not happening. See next entry in this list)
		- Depending on how much local storage someone has available this can help to reduce S3 transfer costs dramatically or if someone does not want that local space to be always filled up reduce that used space. 

- the local copy of S3 files is duplicated for the same song when it plays multiple times in that cleanup timerange, meaning it is not reused when it plays again but a new copy is created locally. See the following excerpt from my local filesystem:
```
-rw-r--r-- 1 azuracast azuracast 6399539 Nov  7 19:00 202011070700359db97_07_-_off_to_see_the_world.mp3
-rw-r--r-- 1 azuracast azuracast 7544713 Nov  7 19:03 202011070702593c22f_01_-_we_got_this_together.mp3
-rw-r--r-- 1 azuracast azuracast 6081970 Nov  7 19:06 202011070706290360f_03_-_time_to_be_awesome.mp3
-rw-r--r-- 1 azuracast azuracast 6783127 Nov  7 19:09 202011070709208e44f_05_-_open_up_your_eyes.mp3
-rw-r--r-- 1 azuracast azuracast 6399539 Nov  7 19:12 202011070712404abf9_07_-_off_to_see_the_world.mp3
-rw-r--r-- 1 azuracast azuracast 7544713 Nov  7 19:15 202011070715411c842_01_-_we_got_this_together.mp3
-rw-r--r-- 1 azuracast azuracast 6081970 Nov  7 19:19 20201107071912c16ec_03_-_time_to_be_awesome.mp3
-rw-r--r-- 1 azuracast azuracast 6399539 Nov  7 19:29 2020110707291695607_07_-_off_to_see_the_world.mp3
-rw-r--r-- 1 azuracast azuracast 7544713 Nov  7 19:31 2020110707314346ba8_01_-_we_got_this_together.mp3
-rw-r--r-- 1 azuracast azuracast 6711237 Nov  7 19:35 2020110707351338574_08_-_thank_you_for_being_a_friend.mp3
-rw-r--r-- 1 azuracast azuracast 6783127 Nov  7 19:38 20201107073828f39d3_05_-_open_up_your_eyes.mp3
-rw-r--r-- 1 azuracast azuracast 6081970 Nov  7 19:41 20201107074148d792f_03_-_time_to_be_awesome.mp3
-rw-r--r-- 1 azuracast azuracast 6399539 Nov  7 19:44 202011070744391e3d8_07_-_off_to_see_the_world.mp3
-rw-r--r-- 1 azuracast azuracast 7544713 Nov  7 19:47 202011070747406823c_01_-_we_got_this_together.mp3
```

- When an S3 StorageLocation is unavailable the playback of songs from that location will fail directly on the next song even if it is cached in the local temp directory. When we get it to re-use already downloaded files those should ideally be still be played instead of switching to the generic offline mp3

- If S3 StorageLocation is unavailable the Media Sync fails with an `S3Exception`
	- If an S3 StorageLocation is unavailable (due to internet outage, routing problems, service is down, whatever...) this should not cause an exception
	- We should skip that storage location and still continue with the other locations
	- What happens to the playlist associations when the Media Sync is running for a station where the StorageLocation is unavailable (after the exception thing is fixed)? Would they be emptied??

- Something with the cache clear is not correctly working (not sure if it's directly related to these changes seeing as there are also some reports about media file processing stuff with SFTP still not working correctly)
	- I've configured my media directory to the S3 bucket you provided with the prefix `vaalyn`
	- Then I've configured my live broadcast recordings directory to the same S3 bucket but with the prefix `vaalyn/recordings`
		- This would enable accessing the recordings through the media files
	- After I've connected, streamed a bit and then disconnected I'd have expected to see the recordings directory and the recording file in the music files browser
		- This is not visible until I manually call the API endpoint with `&flushCache=true`
		- I've tried running the Media Sync task beforehand but this did not seem to correctly clear the cache (the file itself got correctly processed though)

- When backing up a station with S3 Media StorageLocation while NOT excluding the Media from the backup the media files will not be backed up
	- If that is intentional then there should be a warning somewhere that tells the user that media files from S3 StorageLocations will not be backed up

- When using the `Include in On-Demand Player` option for a playlist with files from an S3 StorageLocation these files will not show up on the On-Demand Player page

SlvrEagle23:
@Vaalyn Made a number of improvements to various components, including renaming "Amazon S3" to "S3", simplifying the dropdown selector (I liked your mockup but it's a lot of work that I think is beyond our scope atm), implementing in-form validation for the S3 connection and improving how it handles showing those errors.

I'm okay with "temporary time to keep" being a setting, though I am loathe to create a new setting that only applies to the small subset of S3 users and would likely confuse all the others.

Vaalyn:
👍 will take a look at it after sleeping.

> I'm okay with "temporary time to keep" being a setting, though I am loathe to create a new setting that only applies to the small subset of S3 users and would likely confuse all the others.

How about making it configurable via an environment variable then?

Vaalyn:
## Further issues with previously reported findings

- Validating the credentials for an S3 location does not seem to work
	- Have created an S3 StorageLocation with the credentials you provided and removed the last character of the `Access Key ID` and could successfully save it without seeing any errors

- OnDemand player is working with S3 locations now, one thing I noticed here though is that the file is downloaded directly from S3 without being cached on the server
	- Is that intentional? This could theoretically be abused by malicious people to jack up the S3 transfer costs for a station by repeatedly downloading OnDemand songs from a station with an S3 location.

- When the S3 StorageLocation is not reachable and the song tries to switch to the next one and that one is not cached it will correctly try the next one in the queue and so on and on until it finds one that it can play. That's really good behaviour.
	- Small improvement here might be checking if a song is available before pushing it to the upcoming song queue so that songs that can not be played due to them not being cached won't be queued up and shown on the upcoming song list

- Local StorageLocation path still showing without a leading `/` at the beginning of `var/azuracast/stations/xxxx/xxx`

## Newly identified issues

- Deleting directory on S3 Location says successful but directory is still there after reload

- When moving music files on a S3 Location the `.albumart` and `.waveforms` directories are shown as target locations
	- Should be hidden since users can't access these directories via the music files view
	- Did not check if this is also true on local StorageLocations

- Moving files on S3 Location says it's successful but it does not work and file is still at the same location

- Music files view shows `36.4 MB of Used (6 Files)` which looks rather strange, since there is no way (that I know of) to get the available bucket size of an S3 location this should only say `xx MB used (x Files)`

SlvrEagle23:
@Vaalyn 

> OnDemand player is working with S3 locations now, one thing I noticed here though is that the file is downloaded directly from S3 without being cached on the server...Is that intentional? This could theoretically be abused by malicious people to jack up the S3 transfer costs for a station by repeatedly downloading OnDemand songs from a station with an S3 location.

Currently yes, it's intentional to just stream the response directly from whatever filesystem holds it. That was part of the beauty of the Flysystem construction, that it can just read in a resource from a remote location (like S3) and stream it out through the browser without any real overhead.

Now that you mention it, though, I can definitely see how this could quite possibly rack up immense data egress fees, and since we already have the deterministic "copy to local" thing going on, we should take advantage of that if a local copy already exists for a given item.

I've tried to make sure that elsewhere in the system, data egress is kept to a minimum, i.e. when uploading files the file is processed locally _before_ pushing to S3 so it doesn't then have to be turned back around and downloaded.

SlvrEagle23:
@Vaalyn Everything else should be fixed, but I'm not sure what's going on with the move/rename stuff not working correctly. I think if we can get that figured out, we're otherwise ready to go as far as rolling release goes.

SlvrEagle23:
@Vaalyn Deleting/moving has been fixed thanks to some minor changes in batch handling.

Vaalyn:
> @Vaalyn Deleting/moving has been fixed thanks to some minor changes in batch handling.

👍 can confirm that.

Just one small thing to go then I'd say this is ready to go to the rolling-release:

- The validator for checking if the S3 credentials are working only fails when the `Endpoint` is wrong.
	- When I enter invalid values for `Access Key ID`, `Secret Key`, `Bucket Name` or `Region` the validator doesn't care and happily saves it and tells me everything is fine although it can't access the StorageLocation when I go to Music Files
	- This seems to be because the `getObjectUrl` method of the AWS S3 SDK does not validate the credentials:
		- > The URL returned by this method is not signed nor does it ensure that the bucket and key given to the method exist.
        - I think the best way to check if it is working would be to use the [`listObjectsV2`](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html) method and configure it to only return a maximum of 1 result in order to keep this fast and lean

SlvrEagle23:
@Vaalyn Implemented, and I can confirm this new validation method does check credentials as well.

Vaalyn (approved):

Vaalyn:
> @Vaalyn Implemented, and I can confirm this new validation method does check credentials as well.

✔️ tested it and indeed works

I think you are good to go with this

SC2Mitch:
From my testing, all systems are green, I'll report any issues should any happen when I roll it over to my production instance, let more people get eyes on with the upgrades, great work Silver & Vaalyn.

PR #3349, opened 2020-10-29
Features/expand 'status' api request

Madpeterz:
1. Adds loads and ram to the reply
2. moves __construct() to the top of the class
3. adds 2 new functions (these should be moved out to some where better later)
`systemLoads, systemMemory`

- ### **Untested.**

[related feature request](https://features.azuracast.com/suggestions/129607/expand-status-api-request)

Third Party Content
https://stackoverflow.com/questions/1455379/get-server-ram-with-php/1455610#1455610

### DIFF
--- a/src/Entity/Api/SystemStatus.php
+++ b/src/Entity/Api/SystemStatus.php
@@ -23,7 +23,7 @@ class SystemStatus
     protected function osNotWindows(): bool
     {
         // ref: https://www.php.net/manual/en/function.php-uname.php
-        if (strtoupper(substr(PHP_OS, 0, 3)) === 'WIN')
+        if (strtoupper(substr(PHP_OS, 0, 3)) === 'WIN') 
         {
             return false;
         }
@@ -54,7 +54,7 @@ class SystemStatus
         }
         return [
             "free" => $freeMemory,
-            "used" => $usedMemory
+            "used" => $usedMemory,
         ];
     }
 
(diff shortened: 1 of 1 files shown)


SlvrEagle23:
@Madpeterz Make sure to follow the [PSR-12](https://www.php-fig.org/psr/psr-12/) coding standards, use the array shorthand notation (`[ ]` instead of `array()`), and name variables using `$camelCase`.

Vaalyn (commented):
**Comment about code style:**
Please move the class properties to the beginning of the class since that is how it's done in all other classes in AzuraCast.

**General point that should be discussed:**
In general I'd prefer to not put this type of information in a publicly accessible API endpoint. The `/api/status` endpoint is not secured via a token and the permissions system so anyone would be able to view this information. This is a potential security hazard since having accurate CPU load and RAM usage information can make it easier to run Denial-of-Service attacks on a server by knowing what resources the server has and how it's reacting to requests in regards to resource usage.

This should ideally be added in a new API endpoint that is only accessible when a token of a user with the `view administration` permission is supplied with the request since that is the permission needed to access the `Administration` page in the web interface where this information is visible.

Vaalyn:
Since it is not supported running AzuraCast directly on a Windows server with the native Windows version of PHP I don't think we should add code to specifically check if the installation is running on Windows. When it's running on a Windows server (which officially is only on dev installations) it's running in Docker which is linux based.

Madpeterz:
I would have no problem about having a new api function that is protected todo this.
maybe rename the current SystemStatus to "Heartbeat"

Madpeterz:
Changed this to a draft for now,
going to create a new api function that has the values I want and used the permissions system
while not changing the current api.

Madpeterz:
Removing this pull request for now

Issue #3345, opened 2020-10-28
Swagger is broken in live


Madpeterz:
**Using Docker installation method**
Yes

**AzuraCast version**
v0.10.4 Stable

**Host Operating System**
Ubuntu 20.04 (LTS) x64

**Describe the bug**
after clicking an api option a spinner appears and never goes away for some api items
```
Could not find component JsonSchema_integer_int64 index.js:19:6
Uncaught (in promise) TypeError: e.get(...).keySeq is not a function
    a get-parameter-schema.js:75
    t parameter-row.jsx:101
    t parameter-row.jsx:30
    React 8
get-parameter-schema.js:75:7
    Babel 5
    t _export.js:36
    exports Babel
    Lodash 6
    xe actions.js:243
    $/</</< utils.js:137
    P Redux
    t OperationContainer.jsx:155
    value OperationContainer.jsx:89
    React 7
    w self-hosted:1161
    React 7
    handleChange Redux
    <anonymous> self-hosted:1161
    f Redux
    $/</</< utils.js:140
    P Redux
    u wrap-actions.js:9
    r system.js:174
    z system.js:461
    updateJsonSpec index.js:22
    r system.js:174
    z system.js:461
    ge actions.js:77
    $/</</< utils.js:137
    P Redux
    s wrap-actions.js:5
    r system.js:174
    z system.js:461
    updateSpec index.js:11
    r system.js:174
    z system.js:461
    p download-url.js:37
    (Async: promise callback)
    download download-url.js:26
    $/</</< utils.js:137
    P Redux
    h index.js:153
    Hn index.js:180
    onload index.html:41
    (Async: EventHandlerNonNull)
    <anonymous> index.html:39
```

**To Reproduce**

1. login to your azuracast server
2. goto /api/
3. click a item "/station/{station_id}/backend/{action}"

**Expected behavior**
it to display the api example

**Device(s):**
 - Device: Desktop
 - OS: Windows
 - Browser: Firefox [82.0.1] , Edge [86.0.622.51]

**Additional context**
This might be related
https://github.com/AzuraCast/AzuraCast/issues/3344

SC2Mitch:
This only appears to be happening for backend related things, everything else returns back.

Vaalyn:
@SC2Mitch Here's some debugging info that I got from copy-pasting the content of our `openapi.yml` into https://editor.swagger.io/

```
Semantic error at paths./admin/roles.post.operationId
[...]

PR #3352, opened 2020-10-30
Resolve a issue with frontend/backend via API.

SC2Mitch:
This fixes some of the issues mentioned in #3345 
I believe this should resolve the following: backend{action} | frontend{action} 

Some more work is required for the other issues mentioned but it can be solved at another time.

(Unsure why it removed some things given I didn't touch that part, merely Vaalyn's recommendations in #3345. Feel free to correct / advise if I did something wrong. The fixes was made in https://editor.swagger.io/ and ported over to here. )

### DIFF
--- a/web/static/api/openapi.yml
+++ b/web/static/api/openapi.yml
@@ -9,7 +9,7 @@ info:
 servers:
   -
     url: 'https://demo.azuracast.com/api'
-    description: 'AzuraCast Public Demo Server'
+    description: 'This AzuraCast Installation'
 paths:
   /admin/custom_fields:
     get:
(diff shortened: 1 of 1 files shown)


Vaalyn:
@SC2Mitch Like I mentioned in that issue the `openapi.yml` is generated and thus should not be directly edited.

You need to edit the `@OA\...` annotations, for example in [this block here](https://github.com/AzuraCast/AzuraCast/blob/master/src/Controller/Api/Stations/ServicesController.php#L80-L97) changing:

```
*     content="restart",
*     required=false,
*     @OA\Schema(
*         type="string"
*     )
```

into:

```
*     required=true,
*     @OA\Schema(
*         type="string",
*         default="restart"
*     )
```

and after you've done those changes run the command `./docker.sh cli azuracast:api:docs` to let AzuraCast generate the `openapi.yml` file.

SC2Mitch:
Thanks, I misunderstood what you meant exactly, I'll pull the changes in the morning (I'll double up on coffee, cheers Vaalyn)

PR #3361, opened 2020-11-01
Update for Composer 2.0

Vaalyn:
Since a few days ago Composer 2.0 has been released which means that by installing Composer via the installer from https://getcomposer.org/installer we will automatically use v2.0 when building images (and on ansible installations too I think).

At the moment the `wikimedia/composer-merge-plugin` still requires Composer v1 which is why I've updated it to use the dev branch that is the one that will (most likely) update it to support v2 (locally tested it and it works for me). See their issue regarding Composer v2 compatibility: https://github.com/wikimedia/composer-merge-plugin/issues/184

Also removed the `hirak/prestissimo` composer plugin since it is incompatible with Composer 2.0 which already includes parallel downloads.


SlvrEagle23:
@Vaalyn Tests are failing on this one:

```
Fatal error: Uncaught Error: Interface 'Psr\Http\Message\ServerRequestFactoryInterface' not found in /var/azuracast/www/src/Http/Factory/ServerRequestFactory.php on line 12
#0 /var/azuracast/www/vendor/composer/ClassLoader.php(444): include()

#1 /var/azuracast/www/vendor/composer/ClassLoader.php(322): Composer\Autoload\includeFile()
Error: Interface 'Psr\Http\Message\ServerRequestFactoryInterface' not found in /var/azuracast/www/src/Http/Factory/ServerRequestFactory.php on line 12
#2 [internal function]: Composer\Autoload\ClassLoader->loadClass()

#3 /var/azuracast/www/src/AppFactory.php(54): spl_autoload_call()
Call Stack:
#4 /var/azuracast/www/bin/console(9): App\AppFactory::create()
    0.0001     392064   1. {main}() /var/azuracast/www/bin/console:0
#5 {main}
    0.0228    5946016   2. App\AppFactory::create() /var/azuracast/www/bin/console:9
  thrown in /var/azuracast/www/src/Http/Factory/ServerRequestFactory.php on line 12
    0.0363    7230696   3. spl_autoload_call() /var/azuracast/www/src/AppFactory.php:54
    0.0363    7230760   4. Composer\Autoload\ClassLoader->loadClass() /var/azuracast/www/src/AppFactory.php:54
    0.0363    7230872   5. Composer\Autoload\includeFile() /var/azuracast/www/vendor/composer/ClassLoader.php:322
    0.0364    7235832   6. include('/var/azuracast/www/src/Http/Factory/ServerRequestFactory.php') /var/azuracast/www/vendor/composer/ClassLoader.php:444
```

Likely because of changes Composer 2 has made to the autoloader mechanism, which perhaps some libraries aren't adapted to yet.

Vaalyn:
I might have done something wrong when updating the composer.lock since it also has removed the dependency on `psr/http-factory` which seems to be why this is failing...

Vaalyn:
@SlvrEagle23 I think I found the issue.

Some packages are saying the "provide" some other package which they actually do not do which is why Composer 2.0 does not install that package anymore. (Discussion about this from the Composer repo: https://github.com/composer/composer/issues/9308)

In our dependencies slim/http was doing that which has been fixed in v1.1, see: https://github.com/slimphp/Slim-Http/issues/137

I've also added `psr/http-factory` as a dependency in our main dependencies since we have code that directly references this so we shouldn't rely on other packages requiring it for us.

PR #3444, opened 2020-11-21
This adds a package class to allow for future support of billing systems

xaoseric:
Added a package class to prepare for adding support for billing systems such as whmcs/blesta.

### DIFF
--- a/src/Entity/Package.php
+++ b/src/Entity/Package.php
@@ -1,0 +1,159 @@
+<?php
+
+namespace App\Entity;
+
+use Doctrine\ORM\Mapping as ORM;
+use OpenApi\Annotations as OA;
+use Symfony\Component\Validator\Constraints as Assert;
+use App\Radio\Adapters;
+
+/**
+ * @ORM\Table(name="packages", indexes={
+ *     @ORM\Index(name="idx_name", columns={"name"})
+ * })
+ * @ORM\Entity()
+ * 
+ * @OA\Schema(type="object", schema="Package")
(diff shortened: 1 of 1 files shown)


SlvrEagle23:
@xaoseric It appears this PR only adds a single database entity for "Packages" that is specific to one customer's needs, with absolutely no integration elsewhere in the application. There are no database migrations to actually create the necessary tables, no relationships with the existing tables (and thus these packages don't influence the system at all), no API endpoints to control the data that's in this table, etc.

This should very likely be a custom plugin that's attached on top of AzuraCast, but even then it needs a lot more work before it can be of any use to a customer. We currently have no intention of integrating billing support directly into AzuraCast itself, as our purpose is to provide a free and open-source solution and not to make it easier for users to resell our product while we continue to work as volunteers.

xaoseric:
@SlvrEagle23 would you be against there being a branch with features for hosting providers that acts more as an LTS release for hosting providers with features for packages and allowing resellers?

SlvrEagle23:
@xaoseric We already have a "stable" branch that we'd recommend any third-party resellers use.

It has generally been my philosophy that the AzuraCast core itself should not cater toward resellers, but instead should contain the application code that's universal for all users, particularly the ones who self-host their own installations.

We've built a plugin architecture that lets you hook into menu items, permissions, the database and lots of other things, which is built specifically for custom add-ons used by specific providers. This is definitely where any reseller-oriented code should be built and maintained.

PR #3452, opened 2020-11-23
Add in features for hosting providers

xaoseric:
This adds support for reseller type accounts and features for hosting providers such as packages. So far all the models have been tied in together and the database migration for the additions has been added as well. 

The additions to the station and users tables are nullable so they will not affect the current functionality of the panel. Additionally, the change made to the Acl class will not affect anything either but are there for further integration with the concept of reseller accounts.

I know that you have mentioned before that you are not looking to add hosting features at this time, so I would propose to merge this into a feature/hosting if that is still the case. I do think however, that these additions will benefit hosts looking to integrate with whmcs and other billing systems easier.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4,20 +4,20 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "9589b162fa75cda5e30f9f91a5db6017",
+    "content-hash": "d6b563b65300f7c9f69998e78f2e4502",
     "packages": [
         {
             "name": "aws/aws-sdk-php",
-            "version": "3.158.5",
+            "version": "3.163.2",
             "source": {
                 "type": "git",
                 "url": "https://github.com/aws/aws-sdk-php.git",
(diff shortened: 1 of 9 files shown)


SlvrEagle23:
@xaoseric As mentioned in my response to your previous pull request, I don't find that directly adding functionality whose sole purpose is to facilitate resellers is appropriate for the core of the AzuraCast project itself.

As project maintainers, we've spoken at great length about how we see our interactions with resellers, and this relationship is not one that is necessarily always a positive one for us. A number of resellers have profited heavily from our work without contributing back to us as maintainers, and others still create suboptimal experiences for their users, who mistakenly assume this is what AzuraCast is always like and don't realize they could host it themselves, possibly with better results.

Thus, as an architect of the software one decision I made quite a while ago is that the main AzuraCast project itself is intended to contain only code that serves all users, including self-hosted users. We don't want to be supporting a use-case that doesn't have anything to do with our actual purpose of making web radio accessible to station operators.

However, I do think that these modifications would definitely be well-suited as a custom plugin using our existing plugin architecture. The plugin architecture can override Doctrine entities, permission hooks, etc., and I strongly recommend refactoring this to use that plugin architecture instead.

Issue #3382, opened 2020-11-06
artist & song titles reset during cover upload


InstabilSpielt:
**Using Docker installation method**
Yes

**AzuraCast version**
Rolling Release #02b92bf

**Host Operating System**
Ubuntu 20.04

**Describe the bug**
Since a short time the artist & song title are deleted when I upload a cover.

**To Reproduce**
1. go to music files
2. then click on edit
3. upload album cover

If you try to download the song now, the meta data will be completely reset.

Vaalyn:
Closing as fixed per my PR.

PR #3512, opened 2020-12-07
Fix writing metadata to files via getID3

Vaalyn:
This PR fixes the writing of metadata back to the media files.

The `getid3_writetags` class expects the tags that are set to the `$tag_data` variable to be in the format of:
```
[
    "title" => ["Song Title"]
]
```

Since the implementation of the abstract Metadata read/write interfaces and the `Metadata` class this functionality has been broken due to the values of the tags not being arrays.

I've fixed this in the `GetId3MetadataManager` and not in the `StationMedia::toMetadata` function since this is an implementation detail that is only relevant for the getID3 library.


Vaalyn:
I suspect that this is most likely also related to https://github.com/AzuraCast/AzuraCast/issues/3382

PR #3601, opened 2020-12-24
Fix playback issue in Chromium browsers with Opus streams

### LINKED ISSUE #3597: Opus streams break on song transitions
Yes

**AzuraCast version**
Rolling Release #b970690 (2020-12-23 5:59)

**Host Operating System**

Ubuntu 20.04

**Describe the bug**

When using Opus as the streaming format, and an AutoDJ set up with no crossfading between tracks, the Opus stream stops and fails at the transition.

**To Reproduce**

Create an Opus stream, listen to it until a song transition.

**Expected behavior**

The Opus streams to continue playing between songs like other stream formats. (HE-AAC works fine.)

**Device(s):**
 - Device:  
 - OS: 
 - Browser: 
 - Version: 
Microsoft Edge (the new Chromium version)

**Additional context**

I assumed it might be caused by unusual installation issues, but this a working Docker installation with no modifications, save for some certs I had to copy into the nginx proxy (see #3595). I think it just happens in general with Opus right now, for whatever reason.

These are 24-bit FLAC files, so if that's somehow a factor...

I was hoping to use this as a backend for a Discord bot at some point (including the one being developed by your team?), but this might be an issue there. I assume one workaround would be simply to stream audio continuously rather than relying on the AutoDJ to stitch songs together, though?

Vaalyn:
This PR fixes #3597

At the moment this is not quite ready to be merged since for some reason all other mount points in IceCast are showing `Unkown` for the `Current Song` when I add an Opus mount point. As soon as I disable that one the others show the correct `Current Song` again.

Do you have an idea why this might be @SlvrEagle23 ?

### DIFF
--- a/src/Radio/Backend/Liquidsoap/ConfigWriter.php
+++ b/src/Radio/Backend/Liquidsoap/ConfigWriter.php
@@ -975,6 +975,12 @@ class ConfigWriter implements EventSubscriberInterface
         end
 
         radio = on_metadata(metadata_updated,radio)
+
+        # A metadata-less source of radio for broadcasting OGG streams to prevent premature playback stops
+        # See the following AzuraCast issue: https://github.com/AzuraCast/AzuraCast/issues/3597
+        radio_without_metadata = radio;
+        radio_without_metadata = drop_metadata(merge_tracks(radio_without_metadata))
+        ignore(radio_without_metadata)
         EOF
         );
     }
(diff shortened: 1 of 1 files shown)


SlvrEagle23:
@Vaalyn Weirdly, I'm seeing the exact same issue locally. Even if I remove the `radio_without_metadata` variable entirely and just call that function at the source level (which happens _after_ the other sources are configured to broadcast), as long as that `drop_metadata` function call is in there, it seems to drop the metadata from _every_ version of the `radio` stream, not just the one it assigns to, leaving us with "Unknown" metadata on all mount points.

I have no idea why this is or what we're doing wrong, so we'll need to follow up upstream with the Liquidsoap folks on that one.

Vaalyn:
@SlvrEagle23 after asking on the Liquidsoap Slack about this problem with `drop_metadata` Romain mentioned that he can't reproduce this issue on their `1.4.3` version. I have then tested this on `1.4.3` and `1.4.4-pre-release` and this problem with `drop_metadata` does indeed not exist in that versions. We will have to either wait for the full release of `1.4.4` or use the `1.4.4-pre-release` in order to get this PR working correctly.

Vaalyn:
Closing this PR for now.

PR #3608, opened 2020-12-27
fix(docs): correct link to setup instructions

huyenltnguyen:
This PR adds a small change to correct the link to setup instructions.


Vaalyn (approved):
Good catch 👍

PR #3643, opened 2021-01-06
Fix bug where Music Files list fails to load on S3 storage locations due to missing MTime metadata

### LINKED ISSUE #3640: Music files are there, but won't display

yes

**AzuraCast version**

Rolling Release #991465d (2021-01-05 11:02)

**Host Operating System**

Debian 10

**Describe the bug**

After updating this evening, I am not able to list music files in my station anymore. Media storage location is s3 compatible object storage hosted in Linode which worked perfectly until updating. Up at the top it lists an accurate size and count for the summary, but where the rows were it just shows "Loading..." and spins but never loads the list. Screenshot included. Liquidsoap still works fine, and station plays music. I can even see the folders listed if I click "move". I can see the files listed in playlists (configure playlist as sequential and then "reorder"). Nothing else seems to be affected, only displaying the music files.

**To Reproduce**

1. Update to latest rolling release
2. Go to "manage station"
3. Go to "music files"

**Expected behavior**

Music files should show folders and files in rows.

**Relevant Logs**

I don't see any relevant logs when doing `docker-compose logs -f`. I would welcome an alternative method to see more logs. Log levels are whatever defaults are configured in docker-compose.yml

**Screenshots**

2021-01-06-094545_931x763_scrot

**Device(s):**
 - Device:  Desktop
 - OS:  Arch Linux
 - Browser:  Chromium
 - Version:  Version 87.0.4280.88

**Additional context**

I attempted a couple of things, neither was successful:
1. I restored from a backup taken before the update. Since my storage is in s3, I elect to omit media storage backup, so the backup contained config. The restore works fine, but still won't show media file rows.
2. I redid the storage location config. I just removed it from the station, deleted it from storage locations, then created it new and added it to the station. At first, the summary went back to 0GB 0 files, but then I ran the 5 minute sync from the system debugger page, and it re-imported all the files. I waited for the summary to catch up refreshing music files page every few minutes, but the rows still would not list after it finished.
3. This looks to be like a database problem, but I don't have the knowledge to troubleshoot it.

Vaalyn:
This PR fixes #3640

As I've written in the issue I have no idea when/why an S3 provider returns metadata without a `Last Modified` time in the `HeadObject` API request. The S3 Flysystem Adapter is aware of this however and explicitly checks if this is set in the response so it seems to be something that is accounted for in the Adapter (which basically just means not setting this metadata to the flysystem metadata array).

I'm a bit concerned about this since this could probably also lead to problems with other parts of AzuraCast like for example the `CheckMediaTask` in [this](https://github.com/AzuraCast/AzuraCast/blob/master/src/Sync/Task/CheckMediaTask.php#L209) and [this](https://github.com/AzuraCast/AzuraCast/blob/master/src/Sync/Task/CheckMediaTask.php#L245) line (maybe also other parts where S3 locations are used, like for example backups).

What do you think about this @SlvrEagle23 ?

### DIFF
--- a/src/Controller/Api/Stations/Files/ListAction.php
+++ b/src/Controller/Api/Stations/Files/ListAction.php
@@ -279,7 +279,7 @@ class ListAction
                 }
                 $row->path_short = $shortname;
 
-                $row->timestamp = $meta['timestamp'];
+                $row->timestamp = $meta['timestamp'] ?? 0;
                 $row->size = $meta['size'];
                 $row->is_dir = ('dir' === $meta['type']);
 


krakenfury:
it just so happens that i store automated backups in a folder on the same bucket, and they list just fine in `/admin/backups`.

SlvrEagle23:
@Vaalyn I'm definitely worried about the same thing. If some providers aren't returning modified times for files, then they're probably also flagging constantly for reprocessing. Wondering if we should check for an mtime of zero on files and then apply something like we do for the unprocessable media, where they just reprocess every week or so (with a random time tacked on to avoid stampeding) in case mtimes don't work right.

krakenfury:
awesome @Vaalyn @SlvrEagle23 updated and it's fixed. thank you!

dannyodj:
i have the same problem , but my files are not stored in S3 but in the same server!
im on the lasted AzuraCast • v0.12.4 Stable • Docker • PHP 8.0
how to fix this please .
after loading and loading when there are a bunch of songs i thing 500 songs in one directory it says  : No records to display.  but the folder is full with songs , what to do in this case?

/ my server is runing with 32 gb ram and 2 procesors , i have 15 radio stations there and the cpu load  is 1.77 and 5 gb ram used , should i change the # PHP_FPM_MAX_CHILDREN=5  and # PHP_MEMORY_LIMIT=128M    and # PHP_MAX_EXECUTION_TIME=30 for better song processing perfomance?

i need some help in this case , thank you

PR #3680, opened 2021-01-15
Add Support for Remote Album Art on APIs and Media Uploads

SlvrEagle23:
This pull request creates a new `GetAlbumArt` event that can be hooked into by third-party plugins, along with a first-party implementation of the last.fm API service to pull album artwork when it's not available locally in two situations:
 - When processing song entities in APIs (including the Now Playing APIs), and
 - When uploading new media that doesn't have album art embedded in the file

There is a built-in cache mechanism that stores results for half a day to avoid issues with rate limiting. The album art retrieval can be enabled individually for each of the above scenarios via the system settings panel.


Vaalyn:
I'd like to discuss the refactoring of the `GetId3MetadataManager` to `GetId3Metadata` and the addition of the `MetadataManager`.

As I understand the `MetadataManager` this class is there to ensure the separation of concerns regarding the metadata read / write process in such a way that the pure metadata interaction is still contained in `GetId3Metadata` while integrating 3rd-party metadata services like last.fm into this.

I like this but I'm a bit concerned that `GetId3Metadata` is now a direct dependency of the `MetadataManager` without there being an interface to make it possible to swap out the underlying metadata service.  The `MetadataManagerInterface` is no longer used anywhere but the interface itself does still exists.

I think we should still make use of the interface (but rename it to something `MetadataServiceInterface`) and instead of using `GetId3Metadata $getid3` in the constructor of the `MetadataManager` we should use `MetadataServiceInterface $metadataService` so that the `MetadataManager` is not directly dependent on the `GetId3Metadata` class.

Further more I'd probably also rename `GetId3Metadata` to `GetId3MetadataService` because `GetId3Metadata` sounds like a passive data structure and not like a service that actively does something.

SlvrEagle23:
@Vaalyn Agreed; no good reason to remove the modularity of that interface, so I put it back.

SlvrEagle23:
@Vaalyn Ready for re-review. Now incorporates MusicBrainz and has other important changes to avoid flooding the remote APIs (or over-calling them in the first place).

src/Entity/Song.php
```diff
@@ -90,4 +92,9 @@ public static function createFromText(string $songText): self
         $currentSong = new CurrentSong($songText);
         return self::createFromNowPlayingSong($currentSong);
     }
+
+    public static function offline(): self
```
Vaalyn, line 96: Can we rename this to `createOffline` to match the naming style of the other methods that create an instance of this entity?

src/Event/Media/GetAlbumArt.php
```diff
@@ -0,0 +1,36 @@
...
+    {
+        return $this->song;
+    }
+
+    public function setAlbumArt(?string $albumArt): void
+    {
+        $this->albumArt = !empty($albumArt)
+            ? $albumArt
+            : null;
+        $this->stopPropagation();
```
Vaalyn, line 29: Stopping the propagation here means that after the first handler has called the `setAlbumArt` method no other handlers are called.

Could we explicitly check if `$albumArt` is `null` in order to let it propagate further if one API hasn't found the album art? I know this method is only called when the `!empty($albumArt)` in our handlers is passed so this wouldn't be a problem for us but I'd like to have this designed in a defensive manner just to be safe.

SlvrEagle23:
Merging to let the public play with this feature while working on the next one.

PR #3848, opened 2021-02-27
Implement SMTP Mail Delivery and Self-Service Password Reset

SlvrEagle23:


Vaalyn (approved):
Looks good to me 👍 

Just a small UI nitpick regarding zxcvbn. When I type something into the new password text field it shows a "feedback warning" text that just vanishes again while I'm typing (directly on the next "keyup") which looks a bit confusing.

Additionally I think we should show a password strength indicator based on the `has-error`, `has-success` and `has-warning` classes. The text field border coloration is already a good basis but I'd suggest adding a [`<meter>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meter) element that reflects the scoring and show something like `bad`, `weak`, `strong` as a description for it. This would also increase the accessibility for screen readers by making the strength indicator "visible" to those.

PR #3855, opened 2021-03-02
Improve listener csv export with streamed response

Vaalyn:
This PR further improves the API `ListenerAction` by using a streamed response to download the temporary file instead of loading the whole file into memory and streaming that as a string.

The `renderStreamAsFile` method has a `set_time_limit(600);` in it to have it match what `renderFile` is doing.

Additionally I have separated the code a bit to improve the readability.

Regarding timeouts I think we really need to make the `proxy_read_timeout` more configurable for users. At the moment our `nginx_proxy` is using the default 60s timeout without that being easily overridable. So even if a user increases the php execution time via the env var they will still run into the proxy timeout after 60s.

### DIFF
--- a/src/Controller/Api/Stations/ListenersAction.php
+++ b/src/Controller/Api/Stations/ListenersAction.php
@@ -54,15 +54,6 @@ class ListenersAction
      *   @OA\Response(response=403, description="Access denied"),
      *   security={{"api_key": {}}},
      * )
-     *
-     * @param ServerRequest $request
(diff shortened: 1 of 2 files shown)


SlvrEagle23:
@Vaalyn I'll meet you in the middle on this one; for sure we can do the stream integration and optimize that end of things, but splitting the `ListenersAction` into a bunch of sub-functions that have to pass around local variables like that is a little less appealing to me.

PR #3861, opened 2021-03-06
Chunk switch() statements in Liquidsoap config to avoid hitting limit of scheduled switches

Vaalyn:
This PR fixes a limitation in Liquidsoap's `switch()` statement. When there are too many scheduled switches inside of one statement Liquidsoap can't handle this anymore and (in my case) has been unable to update the streams metadata.

I have chosen the chunk size of 168 to represent 1 scheduled switch per 24 hours for 7 days.
When testing with this size Liquidsoap still behaves normally.

For most users this will still just generate 1 switch statement but for the cases where this size is exceeded it will prevent users from having issues due to that.

### DIFF
--- a/src/Radio/Backend/Liquidsoap/ConfigWriter.php
+++ b/src/Radio/Backend/Liquidsoap/ConfigWriter.php
@@ -428,15 +428,20 @@ class ConfigWriter implements EventSubscriberInterface
         if (!empty($scheduleSwitches)) {
             $scheduleSwitches[] = '({true}, radio)';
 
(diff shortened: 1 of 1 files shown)


SlvrEagle23:
@Vaalyn Shouldn't each resulting chunk have the `({true}, radio)` switch appended to the end of it? May want to move that line of code into the foreach loop so every set will still fall back to the radio signal.

Vaalyn:
Thank you for pointing that out 👍

PR #3956, opened 2021-03-29
Update to Flysystem V2

SlvrEagle23:
This hasn't been fully tested for all use-cases but is passing our test suite and should be in full working order, so it's ready for testing.

Flysystem V2 is a significant change that rips out a fair amount of Flysystem's previous functionality in favor of a locked-down interface without support for plugins, caching, or other important things we depend on, so this rewrite definitely involves some compromises in the way of performance and code quality that I'm not happy with, but which are necessary for us to proceed with a PHP 8.0 upgrade.

Notable changes include:
 - PSR-6 caching is completely absent in Flysystem V2 and is not planned; rather than reimplement it entirely locally, I opted to just keep our generated cache of the output of the `ListFilesAction` and abandon the cache elsewhere. It admittedly did cause us some problems with cache invalidation, but those are issues we still face with our own implementation.
 - As a result of the interfaces of the Local and AWS adapters being privatized and the default Flysystem V2 functionality being to make each individual metadata call a separate API call (even though that API response returns the entire metadata), I had to manually implement a `getMetadata` function on every adapter we use; I hate this, but it's what we need to do to move forward for now. It will mean that should we ever consider adding new adapters in the future, we'll need to do that exact same process for them. Fortunately, since I contributed much of the Dropbox adapter's V2 code, their version remains with protected methods and properties and is vastly easier to extend.
 - Filesystems are classified as either a `LocalFilesystem` or a `RemoteFilesystem` and optimizations are built into the extended filesystem classes for both cases.
 - Since MountManager is similarly privatized and so any extension of it would basically require reimplementing much of its internal functionality, I simply removed it in favor of implementing a `StationFilesystems` class that has built-in methods for pulling individual filesystems. The upside of this is that we no longer need to worry about confusing URIs and paths across the application, since only paths are used.


src/Flysystem/AbstractFilesystem.php
```diff
@@ -0,0 +1,81 @@
...
+        $specialPaths = [
+            '/var/azuracast/backups' => '/internal/backups',
+            '/var/azuracast/stations' => '/internal/stations',
+        ];
+
+        foreach ($specialPaths as $diskPath => $nginxPath) {
+            if (0 === strpos($localPath, $diskPath)) {
+                $accelPath = str_replace($diskPath, $nginxPath, $localPath);
+
+                // Temporary work around, see SlimPHP/Slim#2924
```
Vaalyn, line 71: This SlimPHP issue seems to be resolved: https://github.com/slimphp/Slim/issues/2924 so I think this could be removed.

SlvrEagle23:
Upstream issues with Dropbox are fixed in 2.0.1. Everything tests green on my end!

Vaalyn (approved):
The code changes look good to me. 👍 

I haven't yet thoroughly manually tested it and I will test extending our filesystem from a plugin to add a new station filesystem later on to see if there are any problems with the new flysystem version and our new StationFilesystems.

Since your testing and our automated tests show no issues I would be ok with merging this.

PR #4024, opened 2021-04-15
Fix Typo in German Resources

dgruenberger:
Lider -->Lieder


Vaalyn:
Hey,

translations for AzuraCast are managed through Crowdin.
See https://docs.azuracast.com/en/user-guide/translation

I've gone ahead and updated the text for this translation over there. Next time the translations are updated that change will be included.

PR #4031, opened 2021-04-17
Reimplement playlist queue to be attached to StationPlaylistMedia table.

SlvrEagle23:
This PR moves the queue for playlists from a giant JSON dump in the `StationPlaylist` table to a separate `is_queued` flag in the existing `StationPlaylistMedia` table.

This actually solves a number of problems in our queue management; several other types of problem were solved by returning an API response for the queue, which allows us to pass around a SPM ID that we can use to precisely mark a specific queued item as played (instead of trying to guess it again later).


Vaalyn (approved):
👍 LGTM

Runs fine on my test installation and saw no errors in the logs.

PR #4032, opened 2021-04-17
Add PHP-SPX profiling extension for development environment

Vaalyn:
This PR adds the [PHP-SPX](https://github.com/NoiseByNorthwest/php-spx) extension to AzuraCast which enables us to easily profile and render metrics in a flame-graph like memory usage, io, CPU time, etc. while running in development environment.

The profiler dashboard can be accessed via `http://localhost/?SPX_KEY=dev&SPX_UI_URI=/`

@SlvrEagle23 Do we have a way to only fetch, install and compile the extension for development similar to how some configurations are only included when in development env?
I haven't seen something in the existing scripts that I could copy for this purpose.


SC2Mitch:
While this does look very helpful for development purposes, I propose making it a custom variable in the .env for power users trying to debug stations which are acting abnormally without having to go install additional software, thoughts? (Might be helpful for maintenance / techs to keep a close eye on metrics)

Vaalyn:
> While this does look very helpful for development purposes, I propose making it a custom variable in the .env for power users trying to debug stations which are acting abnormally without having to go install additional software, thoughts? (Might be helpful for maintenance / techs to keep a close eye on metrics)

Good point 🤔 

Maybe something like `ENABLE_PROFILING_EXTENSION=1` to enable it and probably something like `PROFILING_EXTENSION_HTTP_KEY` and `PROFILING_EXTENSION_HTTP_IP_WHITELIST` to control those settings in the `php.ini`?

SC2Mitch:
> > While this does look very helpful for development purposes, I propose making it a custom variable in the .env for power users trying to debug stations which are acting abnormally without having to go install additional software, thoughts? (Might be helpful for maintenance / techs to keep a close eye on metrics)
> 
> Good point 🤔
> 
> Maybe something like `ENABLE_PROFILING_EXTENSION=1` to enable it and probably something like `PROFILING_EXTENSION_HTTP_KEY` and `PROFILING_EXTENSION_HTTP_IP_WHITELIST` to control those settings in the `php.ini`?

That'll work yeah, I do like the whitelist option to ensure only the system admins / techs access this page when they only need to.

PR #4045, opened 2021-04-19
Switch Settings to be a flat single entity to use EntityManager built-in functions.

SlvrEagle23:


src/Entity/Migration/Version20210419043231.php
```diff
@@ -0,0 +1,181 @@
...
+        }
+
+        foreach ($boolFields as $field) {
+            $value = $oldSettings[$field] ?? null;
+            if (null === $value) {
+                $newSettings[$field] = 0;
+                continue;
+            }
+
+            $newSettings[$field] = $field ? 1 : 0;
```
Vaalyn, line 139: While testing on my local dev installation I noticed that it suddenly has `always_use_ssl` turned on which was set to `false` before updating.

This line should probably be `$newSettings[$field] = $value ? 1 : 0;`

SlvrEagle23:
@Vaalyn Made the change requested; everything else looks fine on my installation after quite a while of running. Let me know your thoughts.

Vaalyn:
Also haven‘t noticed any other problems so far. I think it’s good to go live.

Vaalyn (approved):

PR #4160, opened 2021-05-12
Implement podcasting feature

Vaalyn:
This PR intends to implement the long wished for podcasting integration in AzuraCast.

See https://features.azuracast.com/suggestions/51716/podcasting

I have worked on this feature in the form of a plugin in December 2020 and have now started working on integrating it directly into the core of AzuraCast so some parts of the code here might not be up to date with recent changes to AzuraCast yet. 

There are still some things that definitely need to be changed before this can be considered acceptable to merge and I'll list as much of the tasks that are still TODO here in this PR so that we can keep track of those.

I'd love to get some help in identifying things that need to be changed / brought up to code and also with testing as well as general feedback.

## ToDo

- [x] Remove `PodcastCategory` entity in favour of pulling the available categories from the `$availableCategories` property of `MarcW\RssWriter\Extension\Itunes\ItunesChannel`
- [ ] Verify generated RSS feed contains everything needed to correctly list a podcast in:
    - [ ] iTunes
    - [ ] Spotify
- [x] Either make `StationPodcastMedia` entity not tightly coupled to a `Station` & `StationPodcastEpisode` similar to how `StationMedia` works or restrict Podcast `StorageLocation` entities to a single station only
    - **This point needs input. I'm unsure if it even makes sense to have podcast media shared between stations / having the ability to assign podcast media to multiple episodes (in different stations)**
- [x] Verify if any Vue.js component needs to be changed / updated in some way
- [ ] Adapt API endpoints to return OpenAPI compatible API entities

## Resources for testing RSS feed

Apple iTunes Podcast guidelines for the RSS format:
https://help.apple.com/itc/podcasts_connect/#/itcb54353390

Validators:
- https://podba.se/validate/
- https://castfeedvalidator.com/


SlvrEagle23:
@Vaalyn Struggling with some sinus problems today...I'm taking a look through everything and will get back with you periodically, particularly in response to the question about StorageLocation affiliation.

My initial instinct is to treat it exactly the same way we do StationMedia, which is to affiliate it strongly with the StorageLocation and basically have no association with the Station unless that station is specifically incorporating that podcast into, say, a playlist that plays it (which I don't think is even implemented in this first MVP yet).

Of course, `StationMedia` has become something of a misnomer as a result of that change in affiliation (maybe it should now just be...`Media`), so if podcasts are the same way, perhaps they should lose the `Station` prefix unless they're specifically being associated with a specific playlist.

This would allow multiple stations to manage a shared podcast directory by simply having the same storage location, and to then integrate them as needed into playlists (if we build in that feature) while having the permissions remain granular to the per-station level.

SlvrEagle23:
@Vaalyn Also, I noticed that there aren't as many categories listed in the `ItunesChannel` categories as are currently prepopulated in the `PodcastCategory`. Is this intentional, and should the extra categories be coming from somewhere else?

Vaalyn:
> @Vaalyn Also, I noticed that there aren't as many categories listed in the `ItunesChannel` categories as are currently prepopulated in the `PodcastCategory`. Is this intentional, and should the extra categories be coming from somewhere else?

I didn't notice this at first when I included this library but it seems to have not been updated for some time so probably Apple added some new categories since the last update. The ones from the `PodcastCategory` are manually populated by me. I have copied the list of categories from: https://help.apple.com/itc/podcasts_connect/#/itc9267a2f12

It would probably make sense to either fork the library and update it or directly integrate the required classes into AzuraCast.

SlvrEagle23:
@Vaalyn Working on a refactor of a few things to line things up with StorageLocations and also tying the two Category classes together.

SlvrEagle23:
@Vaalyn Latest push switches everything over to StorageLocation, replaces the static category list with a populated array that we can make changes to as code commits, renames classes accordingly and updates the various controllers.

SlvrEagle23:
Finished actually playing with the whole thing all the way down the stack, this should be a working version with various things fixed across the stack.

SC2Mitch:
I'll be free for some testing when this is completed / ready for full reviews, should my availability allow.

SlvrEagle23:
@Vaalyn One question: I'm not seeing a link anywhere from the station-based Podcast management UI to the RSS feed...is that supposed to be in there somewhere?

Vaalyn:
> @Vaalyn One question: I'm not seeing a link anywhere from the station-based Podcast management UI to the RSS feed...is that supposed to be in there somewhere?

I forgot to add one there.

When I built most of the functionality (back when it was a plugin) I was just making sure that the basics are there and working so it can be that some convenience features are missing.

SlvrEagle23:
Just did some big polish passes over the last 2 days to make sure everything's consistent with the rest of the AzuraCast codebase, that any duplicated functionality with other parts of the system use common functions, and other improvements to the UI/UX. Everything's ready for testing now.

SC2Mitch:
Testing ongoing

Vaalyn:
## From testing and playing around:

- We should add a column to the `All Files` tab for podcasts to show that this file is assigned to an episode, making it easier to see what is still unassigned

## From using podcast feed validators

- The cover art must be 1,400–3,000 pixels square to look good on fancy retina-class displays.
    - The podcast art is too small — 500×500 pixels. Apple requires 1400x1400-pixel JPEG cover art.
    - This happens when the image provided is too small or if the default station image is used
- Your feed has no website. We recommend building one so that listeners more easily view and listen to your content.
    - The `channel` -> `link` and `channel` -> `item` -> `link` elements are empty when no website is entered manually, we should default this to the public podcast / episode page when it is empty
- itunes:owner > itunes:email tag is empty.
    - The podcast owner contact information. The <itunes:owner> tag information is for administrative communication about the podcast and isn’t displayed in Apple Podcasts.
- No itunes:image tag in feed. Cannot validate feed artwork.
    - The `$itunesChannel` needs to also have the `$rssImage->getUrl()` set to `$itunesChannel->setImage()`
- Apple changed its podcast categories in July 2019. Currently, you're using the subcategory Video Games under the category Games & Hobbies. This is deprecated with Apple's overhaul and should instead be listed in the category Video Games under the subcategory Leisure.
    - We are currently using the following code to set the ItunesCategories `$itunesChannel->addCategory($category->getTitle());` since the podcast xml library that we are using is slightly outdated the `addCategory()` method matches some categories wrongly. We should either use `setCategories()` with an array as parameter that consists of `string` / `[$parent, $category]` elements or fork the library and correct the outdated parts since the class itself is sadly declared as `final`

SlvrEagle23:
Latest commits add a new default album art image that's 1500x1500 which should also fall within requirements for the podcast aggregators, and also fixes up several of the issues with the RSS feed (including setting default links).

src/Controller/Frontend/PublicPages/PodcastFeedController.php
```diff
@@ -167,20 +175,34 @@ protected function buildRssChannelForPodcast(
...
+    }
 
-            $categories[] = $rssCategory;
-        }
-
-        return $categories;
+    /**
+     * @return mixed[]
+     */
+    protected function buildiTunesCategoriesForPodcast(Podcast $podcast): array
```
Vaalyn, line 194: Could we rename this to `buildItunesCategoriesForPodcast()`?

When I'm reading the current name I'm always reading it as "buildi tunes" instead of "build itunes"

SlvrEagle23:
@Vaalyn The cache busting thing is actually exactly what we do with the station media. The reason we don't just put it in the query parameter is because upstream proxies (i.e. Cloudflare) often ignore the query parameters (or can easily be configured to do so), so we set the modified time directly in the URL and then set the expiration to one year to help optimize the content serving.

Vaalyn:
> @Vaalyn The cache busting thing is actually exactly what we do with the station media. The reason we don't just put it in the query parameter is because upstream proxies (i.e. Cloudflare) often ignore the query parameters (or can easily be configured to do so), so we set the modified time directly in the URL and then set the expiration to one year to help optimize the content serving.

Thanks for the clarification 👍 

I have tested adding the feed URL to the `Podcasts` app on my Mac and it's not showing the artwork.
Fetching the artwork URL for example with `wget` with the URL encoded one from the feed results in a 500 error (haven't found the error message for that one in the logs yet).

SlvrEagle23:
RSS alternates are in the HTML headers of the public pages now.

Not sure how to handle that e-mail contact thing short of adding an extra field to the podcasts, but I would think most people would find it bizarre that they're being asked for an e-mail address in that popup.

Vaalyn:
> RSS alternates are in the HTML headers of the public pages now.
> 
> Not sure how to handle that e-mail contact thing short of adding an extra field to the podcasts, but I would think most people would find it bizarre that they're being asked for an e-mail address in that popup.

👍  validator says the header is working as intended

Regarding the e-mail field I've checked Apples documentation again and the `<itunes:owner>` tag is only listed under `Recommended tags` so I think we can ignore that for now.

Vaalyn:
I have found the exact cause of the issue with the url encoded `|` in the artwork URL and fixed it in the middleware.

One last problem that prevents the feed from being successfully added in Apples Podcast app is the artworks size.
We do already fit the image to a maximum of 3000x3000px but we do not prevent it from being uploaded at a size smaller than 1400x1400px which causes the feed to be rejected completely by the Podcast app.

@SlvrEagle23 what do we want to do, forcefully upscale it to at least 1400x1400 or outright reject uploading images smaller than that?

We'll also need to make sure that artworks pulled from the media files itself conform to that min size.
I think if we fix it in `writePodcastArtwork()` this would take care of both cases (uploaded artwork and pulled from file).

Vaalyn:
Btw. the CI fails because `lstrojny/fxmlrpc` has merged the `dev-php8` branch into their `master` branch ~3 hours ago and the `dev-php8` is not available anymore.

SlvrEagle23:
Updated UX is now ready for testing.

Basically, under the hood the "Episode/Media" separation is kept in place as it is today, but rather than uploading media and separately assigning it to an episode, the episode's media is uploaded right alongside the episode's other metadata (and can be updated via the same modal). This should hopefully yield a far less confusing and more one-to-one relationship between media and podcasts. Also, if the podcast episode has no artwork associated with it, but the media file does, that artwork is automatically applied to the episode (instead of kept separately).

Vaalyn:
I like this re-design of the UX. Much more intuitive in my opinion. 👍 

When adding a new episode to a podcast with a file that has no artwork and while also not supplying an artwork image it does create the episode but also shows an error:
```
TypeError at /var/azuracast/www/src/Entity/Repository/PodcastEpisodeRepository.php L80:</b> App\\Entity\\Repository\\PodcastEpisodeRepository::writeEpisodeArt(): Argument #2 ($rawArtworkString) must be of type string, null given, called in /var/azuracast/www/src/Controller/Api/Stations/PodcastEpisodesController.php on line 235
```

PR #4233, opened 2021-05-29
Implement Doctrine ORM Annotations

SlvrEagle23:
This PR implements the newly-added upstream support from Doctrine ORM for PHP 8.0 attributes. The new ORM version also reflects on the native type hinting of the properties, which means we can finally (with some finessing for IDs) use native property types for Doctrine entities, which really cleans everything up.

There are two DB migrations included:
- The first fixes the discrepancy that has long existed between our entity PHP classes and the DB, where several relational tables allowed null relationships even though in the objects themselves, it was impossible to create an entity without that relationship existing (mostly relationships to `Station`)
- The second finally renames `uid` to `id` for the `User` table so now all auto-increment ID tables use the exact same column name, length and format, allowing for the `HasAutoIncrementId` trait to be used. This migration takes advantage of newer MariaDB's ability to do `ALTER TABLE ... RENAME COLUMN ... TO ...`, which also renames all of the associated foreign key constraints at the same time and avoids a whole host of potential migration issues.

Along with the ORM attribute switch, I also moved things like the AuditLog handler to be attribute-based, and removed the `AuditIdentifier` attribute in favor of classes implementing `Stringable` (which also makes errors in Doctrine itself more helpful, serving a double purpose).

Symfony Validators have also been updated to support attributes, so that migration happened simultaneously, but unfortunately the OpenAPI software has not made the change yet, though moving the docblocks into a "description" annotation does actually yield significantly more helpful OpenAPI spec documents.


Vaalyn:
I have already reviewed the code changes, looking good to me.

Couldn't yet run the branch on my machine to check if I can see any problems.

Currently struggling a bit with Docker / Docker Hub since I need to build a new MariaDB container and for some reason I can't pull the image from Docker Hub:
```
Building mariadb
[+] Building 30.2s (3/3) FINISHED                                                                                                                           
 => [internal] load build definition from Dockerfile                                                                                                   0.1s
 => => transferring dockerfile: 37B                                                                                                                    0.0s
 => [internal] load .dockerignore                                                                                                                      0.1s
 => => transferring context: 2B                                                                                                                        0.0s
 => ERROR [internal] load metadata for docker.io/library/mariadb:10.5-focal                                                                        30.0s
------
 > [internal] load metadata for docker.io/library/mariadb:10.5-focal:
------
failed to solve with frontend dockerfile.v0: failed to create LLB definition: failed to do request: Head https://registry-1.docker.io/v2/library/mariadb/manifests/10.5-focal: proxyconnect tcp: dial tcp 192.168.65.1:3128: i/o timeout
```

Also can't login via the CLI atm due to timeouts so I think Docker Hub could be bugging out:
```
Authenticating with existing credentials...
Login did not succeed, error: Error response from daemon: Get https://registry-1.docker.io/v2/: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
```

Vaalyn (approved):
👍 LGTM

Haven't seen anything broken on my dev machine.

It's a shame that OpenAPI does not yet support Attributes since the mixing of Attributes and Annotations does look a bit ugly but I'm sure with time they'll also support them and we can switch those over too.

PR #4395, opened 2021-07-11
Add AutoDJ setting for looping playlist only once when scheduled

Vaalyn:
**Fixes issue:**
See https://github.com/AzuraCast/AzuraCast/discussions/4186

**Proposed changes:**
This PR adds an option to each schedule entry of a playlist that makes it possible to control if the playlist should only loop through its contents once at the scheduled time. When the next scheduled time for a playlist is due it will also check if the playback queue of the playlist if completely full and if not reset it accordingly to make sure that the playlist is played from the beginning.

### DIFF
--- a/src/Radio/AutoDJ/Scheduler.php
+++ b/src/Radio/AutoDJ/Scheduler.php
@@ -2,7 +2,10 @@
 
 namespace App\Radio\AutoDJ;
 
+use App\Doctrine\ReloadableEntityManagerInterface;
 use App\Entity;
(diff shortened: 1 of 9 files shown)


src/Radio/AutoDJ/Scheduler.php
```diff
@@ -12,7 +15,10 @@
 class Scheduler
 {
     public function __construct(
-        protected Logger $logger
+        protected Logger $logger,
```
SlvrEagle23, line 18: Should probably inject LoggerInterface to be consistent with the rest of the app.

PR #4405, opened 2021-07-16
System-Wide Strict Types

SlvrEagle23:
This PR updates the entire main portion of our application to have the `declare(strict_types=1)` header present.

One of the main "trouble points" that this brought up was the fact that `App\Http\RouterInterface` returned a `UriInterface` for one of its functions but a `string` for the others. As part of this PR, I've updated this to be uniform in returning a `UriInterface`, which requires casting to a string in many cases where it's returned directly or processed in a JSON response.

Thanks to the improvements (and upstream improvements with libraries) we're now able to bump up to PhpStan level 5 testing (level 6 is vastly more complex and throws errors everywhere).

Our test suite is green on this version and my local tests show no problems at this point, so we just need to test things out with this version and ensure everything's working as intended.


Vaalyn (approved):
Looks good to me 👍 

Haven't encountered any issues on my test installation and no errors in the log.

PR #4418, opened 2021-07-20
Switch message queue from MariaDB to beanstalkd

Vaalyn:
**Proposed changes:**
This PR replaces the currently used `DoctrineTransport` for the Symfony Messenger with the `BeanstalkdTransport` and installs the respective `beanstalkd` message queue service.

### ToDo List:
- [x] Install beanstalkd
  - [x] Docker
  - [x] Ansible
- [x] Starting of the beanstalkd service
  - [x] Docker
  - [x] Ansible
- [x] Removing `MessengerMessage` Doctrine entity and respective table
- [x] Testing


SlvrEagle23:
Figured out why testing was having trouble...beanstalkd doesn't start up when you're running `docker-compose run --rm` on the web container. To work around this, for testing purposes we have to use in-memory transports to accomplish the same purpose.

To switch out between the two I did the usual pattern, `QueueManagerInterface` with common methods, an `AbstractQueueManager` with shared methods/props and a new `TestQueueManager` class that's solely in-memory and never calls Pheanstalk.

That green-lights all of our tests with no issues.

All of my local testing, including reviewing things via the Aurora UI (very handy tool if you haven't played with it), has shown it's a very capable and very _fast_ little message queue. Should be an excellent option for us.

Give it another look through and let me know how it looks and we'll get it merged in.

Vaalyn:
Looking good to me on my test installation 👍 

Do we want to have this merged and tested for some time before getting back to enabling multiple workers again or should we also enabled it now to get immediate feedback about it?

PR #4595, opened 2021-09-20
Frontend Security with Backend Administration Access

aarkkappler:
Hi 

I would be great for security issues wp do, to only give the admin access and login access for the first installation process, and only offer a front-page when AzuraCast is set as frontend

The Two-Factors Auth Built-In is a VERY good point, already


SlvrEagle23:
@aarkkappler Please stop submitting your discussions or ideas in the form of Pull Requests.

Pull requests are when you're contributing actual changed code to the application, not just discussing an idea.

PR #4845, opened 2021-12-07
fix: imported playlist's filenames are now sanitized

hcbd:
Uploaded files have their filename sanitized (only checked the webbrowser upload, not sure about sftp uploads). 
When importing a playlist from the same source the filenames do not match if they contain uppercase or spaces in their filepaths / filenames.

**Proposed changes:**
Simple addition to sanitize the filename part of the imported playlist.

`// uploaded filenames are sanitized (tolowercase and space to _)
// maybe of code below
$path_parts[$path_parts_length - 1] = strtolower( 
    str_replace(' ', '_', $path_parts[$path_parts_length - 1]) 
);`

I also found a function that does the sanitizing; Class UploadedFile->filterOriginalFilename(filename) instead 
but not knowing the azuracast api as a whole, i don't know if it will have side effects.

This should only sanitize the filename part of the path, because a folder can contain spaces and uppercase letters (within Azuracast).

Please add any improvements or fixes you may spot.

### DIFF
--- a/src/Controller/Api/Stations/Playlists/ImportAction.php
+++ b/src/Controller/Api/Stations/Playlists/ImportAction.php
@@ -70,14 +70,31 @@ class ImportAction extends AbstractPlaylistsAction
                 // De-Windows paths (if applicable)
                 $path_raw = str_replace('\\', '/', $path_raw);
 
-                // Work backwards from the basename to try to find matches.
+                // split path into parts
                 $path_parts = explode('/', $path_raw);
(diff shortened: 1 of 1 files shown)


SlvrEagle23:
@hcbd This is definitely a valid issue, but I wanted to take a different approach, by sanitizing both the existing files and the incoming playlist paths, so that they're a like-for-like comparison whether or not they were uploaded via the web uploader. Closing this in favor of the most recent commit, which strengthens playlist importing quite a bit.

PR #4894, opened 2021-12-18
Feature: Playbutton in playlist re-order modal

hcbd:
**Feature:**

Frontend: A playbutton for songs in the re-order playlist modal

**Proposed changes:**

- Add playbutton before song title in table in re-order modal
- Add inlineplayer to top of modal, so the playback can be controlled and is not hidden behind the modal in the header

Feel free to comment with any improvements or suggestions; 
i'm guessing there is a better way of getting the station_id + hostname inside this component (see comment in code)

Thank you for your time.

**Preview modal:**
afbeelding


SlvrEagle23:
@hcbd Excellent UI implementation, though I'd muuuch rather just add the download URL into the API response instead of doing the hacky thing. Will look into implementing that.

hcbd (commented):
This seems like a better solution than my hacky way of getting the api url

PR #4923, opened 2021-12-25
Add gulp watch script

stathismor:
Add gulp watch script to watch for vue and js file changes

**Fixes issue:**

I want to have the frontend automatically be built when I change vue and js files so frontend development is sped up.

**Proposed changes:**
1. Added a new `gulp watch` script
2. It could be part of the default gulp run, but didn't want to change existing dev process too much
3. Only looks at `vue` and `js` files and it rebuilds **everything**. This can be further optimised but this is a start.
4. Renamed `bundle_deps` to `bundle-deps` to align with existing naming convention
5. After changes have been built, you still need to refresh the page on your browser to see them. I haven't looked at how this part can also be automated (maybe an apache config?)

### DIFF
--- a/frontend/gulpfile.js
+++ b/frontend/gulpfile.js
@@ -84,7 +84,7 @@ defaultTasks.forEach(function (libName) {
   });
 });
 
-gulp.task('bundle_deps', gulp.parallel(
+gulp.task('bundle-deps', gulp.parallel(
   defaultTasks.map(function (name) {
     return 'scripts:' + name;
   })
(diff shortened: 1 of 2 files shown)


pixel-paul:
@stathismor - for point 5:  maybe https://browsersync.io would be appropriate?

stathismor:
@pixel-paul I haven't looked at this again yet. I _think_ it's not a node server that serves the public page and I believe I saw apache somewhere, that's why I mentioned it above... There is an elaborate build/bundle process (gulp+webpack+other stuff) that puts files in a `web` folder, but not sure what picks it up from there.

Basically, I want to build my own player page. I am trying to decouple the public page from the backend and other front-end stuff (like admin), but this is quite tricky. It would be amazing if the public page was independent....

pixel-paul:
Browsersync is great as you can add it as a gulp task, so that after the scripts are compiled, browsersync will reload the browser.

My approach for the player was to make it completely independent of Azuracast (and hosted externally to the Azuracast instance), using the API to get all the info I need.

stathismor:
>  Browsersync is great as you can add it as a gulp task, so that after the scripts are compiled, browsersync will reload the browser.

@pixel-paul all right, I thought it might need a node server to run, to force-refresh the browser. Sounds like it might work then. 

>  My approach for the player was to make it completely independent of Azuracast (and hosted externally to the Azuracast instance), using the API to get all the info I need.

Cool! So, have you done that? My front-end skills are not that strong, so I trying to re-use existing Vue and CSS.

pixel-paul:
> Cool! So, have you done that? My front-end skills are not that strong, so I trying to re-use existing Vue and CSS.

As we had migrated from Centovacast, we already had a player as a starting point. We use Howler JS (https://github.com/goldfire/howler.js) for the player, jQuery 3 for the DOM manipulation and loading data from the API, then Bootstrap 5 as the CSS framework. Aim is to keep it as simple as possible. 

 I have started working on a VueJS version, but have not had the opportunity to develop it more than a PoC.

pixel-paul:
Oh and missed this important step, to connect to the API from Azuracast (using PHP):

https://github.com/AzuraCast/php-api-client

stathismor:
Nice, thanks @pixel-paul, if I ever have something working, I will post it here :smile:

pixel-paul:
Just realised we're commenting about this on the PR. Best to post to a discussion!

PR #4941, opened 2022-01-03
Implement Concurrent Optimized Sync and Now Playing Processes

SlvrEagle23:
This commit fully overhauls the routine synchronization and "Now Playing" sync process to optimize for performance and reliability.

Changes include:
 - The 1-minute, 5-minute and 1-hour tasks have all been replaced by a single task that runs every minute and checks whether tasks are scheduled to run on that minute.
 - Each task is then configured with a cron-style pattern for how it repeats, i.e. once every minute, or 5 minutes, or at the 23rd minute of the hour; this allows us to "stagger" the tasks across the hour instead of having all the longer ones run at exactly the top of the hour, right on top of the 1 and 5-minute tasks.
 - Every task is run concurrently in its own isolated process and managed by an "overseer" PHP process, which manages the exclusivity locks and other things for the process, so even if the task itself crashes, it only affects the single task and won't hold a lock open forever.
 - The Now Playing sync task is now an ongoing PHP worker process that, every 1.5 seconds, checks all active stations and ensures they've recently pinged for metadata within the last 5-15 seconds. The ping time is slightly randomized to avoid stacking stations together repeatedly. Each station's now-playing check is also process-isolated to avoid an issue with one station causing others to fail.

So far in my tests, everything works wonderfully with this new setup, and it looks like this sync process should be _significantly_ more performant, fault-resistant, and capable of handling a huge volume of stations before encountering the kinds of "race conditions" that have plagued large AzuraCast installations before.


SC2Mitch:
I'll get around to testing this tomorrow or Tuesday by latest.

PR #4977, opened 2022-01-11
Enhancement/ansible

aurelienmaury:
Hi!

I am really into ansible code writing and would like to contribute on this part. My first proposition is this PR with many small improvements to code quality, guided by ansible-lint (2h of work).

If you wish to discuss about it, I'm open.

Thanks for the great work. Hope that'll help.

### DIFF
--- a/util/ansible/.ansible-lint
+++ b/util/ansible/.ansible-lint
@@ -74,4 +74,4 @@ extra_vars:
   cert_src: lint
   cert_dst: lint
 
-# Finished with 175 failure(s), 21 warning(s) on 66 files.
+# Finished with 97 failure(s), 21 warning(s) on 66 files.
--- a/util/ansible/.ansible-lint
+++ b/util/ansible/.ansible-lint
@@ -74,4 +74,4 @@ extra_vars:
   cert_src: lint
   cert_dst: lint
 
-# Finished with 97 failure(s), 21 warning(s) on 66 files.
+# Finished with 18 failure(s), 9 warning(s) on 66 files.
--- a/util/ansible/.ansible-lint
+++ b/util/ansible/.ansible-lint
@@ -74,4 +74,4 @@ extra_vars:
   cert_src: lint
   cert_dst: lint
 
-# Finished with 265 failure(s), 21 warning(s) on 66 files.
+# Finished with 175 failure(s), 21 warning(s) on 66 files.
--- a/util/ansible/.ansible-lint
+++ b/util/ansible/.ansible-lint
@@ -74,4 +74,4 @@ extra_vars:
   cert_src: lint
   cert_dst: lint
 
-# Finished with 300 failure(s), 21 warning(s) on 66 files.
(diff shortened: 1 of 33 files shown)


aurelienmaury:
I don't know how much time the CLA Assistant takes to refresh but be sure that I "have agreed to the CLA for multiple repositories or organizations" (Version: 2020-07-14).

Regards.

SC2Mitch:
I'll be getting around to look at this tonight or tomorrow

util/ansible/.ansible-lint
```diff
@@ -0,0 +1,77 @@
...
+
+# Define required Ansible's variables to satisfy syntax check
+extra_vars:
+  scope: lint
+  workspace: lint
+  receiver: lint
+  cert_src: lint
+  cert_dst: lint
+
+# Finished with 18 failure(s), 9 warning(s) on 66 files.
```
SC2Mitch, line 77: Is this included for any specific reason, or left here accidentally?

SlvrEagle23:
Hello! Thanks for figuring out the CLA thing.

Looking through this, I do appreciate the Ansible linting cleanup; it does make the files look a lot more consistent and easier to maintain. I'm completely happy to merge all of those in.

There are some other supporting files, though, that I don't think are necessary at all for our Ansible functioning (like `.envrc`) and were added in this PR. I don't see why we need those and I would be more interested in merging the PR if it was just the linting fixes.

util/ansible/.ansible-lint
aurelienmaury: For the initial question: That is on purpose. That's the result of an `ansible-lint` run associated with the code in the commit. I made this line evolve with each commit so you can monitor the impact. Glad you noticed. :-)

Building on this: `cert_*` occurence where useless so I cleaned up the `.ansible-lint` file so only useful configurations keys are present and ref doc is listed.

aurelienmaury:
Sorry for that. I deleted `.envrc` and `Makefile`. I will come back later on this but with documentation and in another PR. ;-)

SlvrEagle23:
@aurelienmaury There's still a few other files that were added, would you mind removing those too just for this PR?

aurelienmaury:
I guess that should be fine now. Tell me if I missed anything.

Vaalyn:
@aurelienmaury I've got a small question that is not directly related to this PR but since you are quite knowledgeable with Ansible I figured it might be worth asking here.

I just saw a GH issue about a problem with our `Remove Old PHP Libraries` Ansible task were the step fails due to the following error: `No package(s) matching 'php5-*' available`.

When researching about this I didn't really find anything useful about how to prevent the Ansible Package Manager module from failing when a package is unavailable in general. Do you have an idea how to do this?

aurelienmaury:
I think you can't prevent that behavior (or by ugly means like relying on the `shell` module, which moral disapproves).

The key here is to avoid falling in the case by relying on the gathered facts (aka Hardcore Variabilization) to handle specific cases depending on distribution, version and architecture.

That step will be my next enhancement proposal.

PR #4987, opened 2022-01-15
Enhancement/ansible support

aurelienmaury:
That refactoring is aimed at applying a code structure that takes system architecture, distribution and version in account in a low effort way.

I began with the `init` role.

That code was tested on Ubuntu 16.04 & 18.04 & 20.04 and playbooks should run as-is without any modification.


aurelienmaury:
I just saw [here](https://docs.azuracast.com/en/getting-started/installation/ansible) that you support only Ubuntu LTS 18 and 20... So can I drop any Ubuntu 16-specific legacy code?

SC2Mitch:
I believe that since 16 is EOL, it might be best to work towards removing any legacy code relating to Ubuntu 16, but thats my opinion of it.

SlvrEagle23:
@aurelienmaury That is correct, you can remove any Ubuntu 16.04-specific code in Ansible, especially any code paths specifically dedicated to that version that differ from other versions. We follow the Ubuntu software support lifecycle, so only 18.04 and 20.04 are currently supported, and we will hope to have 22.04 supported very soon after it launches.

Issue #5002, opened 2022-01-18
Listeners page returns incorrect data - Internal/Reserve IP location


simpsss:
### Installation Method

Docker Installation

### AzuraCast Release Channel

Rolling Release Channel

### Current AzuraCast Version

#dcea1b9

### What happened?

After the last update (#dcea1b9), is not showing the real IPs and locations of the people who are listening.

Maybe it's because you changed the nginx proxy?

![image](https://user-images.githubusercontent.com/51731449/150011494-c81a0de7-7d7a-4d5a-89a3-6561666f91f1.png)

SC2Mitch:
Shall check in a moment, standby.

dvdred:
> I have the same issue on my installation on docker, version #fcdc50c
I don't know why, but i solved the issue in this way:
- docker-compose down of azuracast
- update and reboot of the docker host (CenOS 7)
- docker-compose up of azuracast (with the same version: #fcdc50c)
Now i can see the public ip of the listeners.

SlvrEagle23:
I suspect this may be indeed caused by our retirement of the multi-container nginx proxy solution, but it should be fixed after updating.

Some possible causes of problems may be:
 - Do you have a `docker-compose.override.yml` file? Make sure that's updated to not have any references to `nginx_proxy` or `nginx_proxy_letsencrypt_companion` (unless you're actually using a multi-site configuration)
 - Make sure when updating that you run `./docker.sh update-self` first, to ensure the updater is running all the necessary steps

SC2Mitch:
Updated to latest rolling, and still retains the Internal IP
![image](https://user-images.githubusercontent.com/29955942/150539654-8c284c77-10e7-4f5e-99e6-a4a29dd18a93.png)

Not using a override file.

SlvrEagle23:
@SC2Mitch And just to confirm, there's no `nginx_proxy` service listed in the main `docker-compose.yml` file?

SC2Mitch:
Nope, this is my main file for clarity
```
services:
    web:
        container_name: azuracast_web
        image: 'ghcr.io/azuracast/web:${AZURACAST_VERSION:-latest}'
        ports:
            - '${AZURACAST_HTTP_PORT:-80}:80'
            - '${AZURACAST_HTTPS_PORT:-443}:443'
            - '${AZURACAST_SFTP_PORT:-2022}:2022'
        depends_on:
            - mariadb
            - stations
            - redis
        env_file: azuracast.env
[...]

PR #5026, opened 2022-01-23
Fix 5002 Internal IP location

YosefAdPro:
**Fixes issue:**

#5002 
**Proposed changes:**

### DIFF
--- a/util/docker/web/nginx/azuracast.conf.tmpl
+++ b/util/docker/web/nginx/azuracast.conf.tmpl
@@ -169,6 +169,8 @@ server {
 
         proxy_set_header Host localhost:$1;
         proxy_pass http://stations:$1/$3?$args;
+        proxy_set_header X-Real-IP $remote_addr;
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
     }
 
     # Reverse proxy the Liquidsoap harbor inputs to allow for streaming.


SlvrEagle23:
I'm not sure why this actually fixes it. The proxy settings are defined higher up and should be pre-empting any of those settings...there should be no reason that redefining it farther down fixes it. 🤔

YosefAdPro:
This may be because it is outside the parameter  `server {}`?

YosefAdPro:
This may be because it is outside the parameter `server {}`?

Because when I put these two lines  into `location ~ ^/radio/` I got the  Public IP

Vaalyn:
@SlvrEagle23 I think I know why it fixes the issue when it's set in the block like in this PR:

> The proxy_set_header directives are inherited from the previous level if and only if there are no proxy_set_header directives defined on the current level.

http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_set_header

We have a `proxy_set_header Host localhost:$1;` in that block already so the previously defined one is not inherited.

SlvrEagle23:
> @SlvrEagle23 I think I know why it fixes the issue when it's set in the block like in this PR:
> 
> > The proxy_set_header directives are inherited from the previous level if and only if there are no proxy_set_header directives defined on the current level.
> 
> http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_set_header
> 
> We have a `proxy_set_header Host localhost:$1;` in that block already so the previously defined one is not inherited.

Well THAT'S frustrating, because we only need that `set-header` in that one block...

PR #5086, opened 2022-02-09
Only append sid to PW for SHOUTcast v2 frontends

### LINKED ISSUE #5085: LS Remote relay password have :#n added to the end of them ( SHOUTcast V1 )
### Installation Method

Docker Installation

### AzuraCast Release Channel

Stable Channel

### Current AzuraCast Version

0.15.0

### What happened?

Shoutcast V1 remote relay passwords have stream_id added to the end of the string `:#N` see test config below

```
# Remote Relays
output.icecast(%mp3(samplerate=44100, stereo=true, bitrate=320, id3v2=true), id="relay_1", host = "Remote URL", port = portNum, password = "somePassWord:#1", name = "Radio URL", description = "Main Station", genre = "charts", url = "Radio URL", public = false, encoding = "UTF-8", protocol="icy", radio)
```
*info was removed from the config above ( port URL's and Passwords )

### Relevant log output

```shell
2022/02/09 16:09:44 [relay_1:3] Connecting mount sid#1 for @Remote URL...
2022/02/09 16:09:44 [relay_1:2] Connection failed: bad answer: invalid password
2022/02/09 16:09:44 [relay_1:3] Will try again in 3.00 sec.
```

Vaalyn:
**Fixes issue:**
Fixes #5085 

**Proposed changes:**
The `:#N` part in the password is only used by SHOUTcast v2 although SC v1 also uses the Icy protocol.
This PR makes sure that this `sid` is only added to SC v2 mounts.

Since the [Entity\StationRemote::getAutodjPassword()](https://github.com/AzuraCast/AzuraCast/blob/main/src/Entity/StationRemote.php#L196) already returns the password with the remotes configured `sid` appended I have switched this part to only add it when it's a local SC mount.

I originally wanted to adapt the [Entity\StationMount::getAutodjPassword()](https://github.com/AzuraCast/AzuraCast/blob/main/src/Entity/StationMount.php#L386) to match the behaviour of the one from the `StationRemote` but noticed that for local SC 2 frontends the `sid` is generated by a manual index number.

This brought up question for me.
A bit further down there is this block of code:

```php
if (!empty($mount->getAutodjMount())) {
    if (StreamProtocols::Icy === $protocol) {
        $output_params[] = 'icy_id = ' . $id;
    } else {
        $output_params[] = 'mount = "' . self::cleanUpString($mount->getAutodjMount()) . '"';
    }
}
```

For remote SHOUTcast v2 mounts this will not add the `icy_id` output parameter to the output in Liquidsoap.
This means Liquidsoap will automatically use the default `icy_id=1`. In those cases this should probably also be set to the respective `sid` as configured in the remote mount. I'm not sure what unintended side effects having this always be `1` for remote SC 2 mounts could have.

@SlvrEagle23 What do you think about this?

### DIFF
--- a/src/Radio/Backend/Liquidsoap/ConfigWriter.php
+++ b/src/Radio/Backend/Liquidsoap/ConfigWriter.php
@@ -1057,15 +1057,17 @@ class ConfigWriter implements EventSubscriberInterface
         if (!empty($username)) {
             $output_params[] = 'user = "' . self::cleanUpString($username) . '"';
         }
-
-        $protocol = $mount->getAutodjProtocolEnum();
-
+        
         $password = self::cleanUpString($mount->getAutodjPassword());
-        if (StreamProtocols::Icy === $protocol) {
+        
+        $adapterType = $mount->getAutodjAdapterTypeEnum();
+        if (FrontendAdapters::Shoutcast === $adapterType) {
             $password .= ':#' . $id;
         }
+        
         $output_params[] = 'password = "' . $password . '"';
 
+        $protocol = $mount->getAutodjProtocolEnum();        
         if (!empty($mount->getAutodjMount())) {
             if (StreamProtocols::Icy === $protocol) {
                 $output_params[] = 'icy_id = ' . $id;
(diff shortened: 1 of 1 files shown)


SlvrEagle23:
@Vaalyn I'm not sure what (if any) the impact of this is. If it's something that really should just be isolated to Shoutcast, then yeah, that logic should be changed...I have no recollection of why it was this way in the first place.

PR #5167, opened 2022-03-04
AzuraCast in a Single Docker Image

SlvrEagle23:
This pull request consolidates all of our remaining Docker containers (web, mariadb, redis and stations) into a single container based around the shared foundation that the `stations` and `web` containers already had.

### Benefits

 - Third-party software like Portainer can now deploy and manage an AzuraCast instance by deploying a single image
 - Reduced image sizes due to greatly reduced redundancy in configuration

### Disadvantages

 - Currently, MariaDB and Redis can't be disabled in part or wholesale from running inside this container. This means if AzuraCast is running in a distributed environment with other applications and the user would prefer to use an external MariaDB database, while the _application_ itself will allow this via environment variables, the MariaDB service would remain running inside the container despite not being used; the same applies with Redis, but given Redis's much smaller footprint, this is much less of an issue.
 - The CI test suite will need to be rewritten to take advantage of the existing `my_init` function to spin up services that the application depends on, instead of counting on those services being available elsewhere via Docker Compose.

### Technical Notes

 - The combined image is built on top of the MariaDB Ubuntu Focal image because the MariaDB installation involves the most customized variant of the upstream software, including their own special scripts for initial startup and maintenance. Other than MariaDB-related changes, this is a very barebones Ubuntu image, so it makes a good candidate to build on top of.
 - The build process has been slightly modified to accommodate this much larger build, so that each previous container's worth of tasks are bunched into a single cacheable layer, with the most computationally intense of the layers appearing before the others in the build process. This means that as long as our `stations` configuration hasn't changed, the CI builds will simply use the full cached image of that portion, saving immensely on build times.
 - The initial implementation of this system uses `runit` as both the stations and web containers already used it and it's a very simple process manager, but we should strongly look into using supervisord, which is now present in the image as well (managing the `stations` components) to manage _all_ of the processes in the container. If nginx, php-fpm, etc. are _themselves_ managed by Supervisord, it would allow us to trigger reloads or restarts of those services from within the application itself, something that's not currently possible.


SlvrEagle23:
Updates made to the image since the original PR:

 - MariaDB and Redis services in the unified container will automatically disable themselves if they detect that their respective `MYSQL_HOST` or `REDIS_HOST` environment variables aren't set to `localhost`. The rest of the system is capable of starting up with or without those services present.
 - Startup scripts have been separated into two pools, `minimal` and `full`, with minimal being the set required for any CLI functionality, CI tests, etc. and full being the typical full web stack running on its own.
 - The init script has been modified to auto-detect whether it's running with or without a command present, and if it's running with a command, it automatically spins up the "minimal" pool only, waits for it to be ready and then runs the command; without a command, it spins up the full stack and monitors that instead.
 - The CI is now fixed and will use the standalone image system without any issues, as will any local commands executed with `docker-compose run --rm web`.
 - A number of other bug fixes and optimizations around the system relating to the new "standalone" Docker type.

At this point, I've tested this locally through numerous iterations and I'm reasonably confident it's ready for broader testing and will soon be ready for release.

Vaalyn (approved):
Looks good to me 👍 

Encountered no issues on my test server so far.

### Posting this here in addition to Discord too:

One thing we should do before merging this is looking through the docs for any tutorials / commands that reference one of the now non-existent containers and put a second version of that specific thing in there. One for the old way and one for the new way in order to 1st have them still be there for users of the older versions and 2nd to have users of the latest RR versions not confused by the "missing" containers. When we have a stable version released with this change included we could think about removing those "duplicates" from the docs again.

SC2Mitch:
I'll get to work on changing the docs to add disclaimers that it'll be changing, will handle this evening.

PR #5308, opened 2022-04-23
Add additional processable mime types

Vaalyn:
**Feature Request:**
https://features.azuracast.com/suggestions/198622/new-file-formats-allowed

**Proposed changes:**
This PR is adding some additional mime types to the list of processable mime types.

Thanks to the new fallback to FFmpeg in our metadata manager when a file that getID3 does not support we can expand the list of processable media types a lot.

For this I'd like to gather some feedback from users to see what media formats they'd like to be able to upload and thus use for their radio stations.

## New Supported Formats:
- `.stm` / `.s3m` (ScreamTracker Module / ScreamTracker 3 Module)
  - https://en.wikipedia.org/wiki/Scream_Tracker
  - https://en.wikipedia.org/wiki/S3M_(file_format)
- `.mod` / `.xm` (Module / Extended Module)
  - https://en.wikipedia.org/wiki/MOD_(file_format)
  - https://en.wikipedia.org/wiki/XM_(file_format)
- `.aiff`
  - https://de.wikipedia.org/wiki/Audio_Interchange_File_Format
- `.wma` / `.wmv` / `.asf`
  - https://de.wikipedia.org/wiki/Advanced_Streaming_Format
  - https://de.wikipedia.org/wiki/Windows_Media_Audio
  - https://de.wikipedia.org/wiki/Windows_Media_Video

## For Users:

If you'd like to have a specific media format added to this list please post the following information.

**Required**
- Name and typical file ending (for example name `MP3`, file ending `.mp3`)

**Nice to have**
- Mime Type (for example `audio/mpeg`)
- Link to a page that explains the format (if something more exotic)
- Example files (so that we can test them)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,8 @@ release channel, you can take advantage of these new features and fixes.
 
 - You can now specify a custom max timeout for "Generic" web hooks (that make HTTP requests to external URLs).
 
+- You can now add Storage Locations that use SFTP connections.
+
 ## Code Quality/Technical Changes
 
 - **Unified Docker Container**: We have combined all of our Docker containers into a single unified container that
@@ -58,6 +60,10 @@ release channel, you can take advantage of these new features and fixes.
 
 - Internal API requests (Icecast listener auth, Liquidsoap API calls) are now handled via a separate, dedicated "back
   channel" and won't be affected as severely by heavy traffic on AzuraCast from public viewers or administrators.
+
+- When processing media, we now use a combination of two libraries (php-getid3 and ffmpeg/ffprobe) to process media and
+  retrieve rich metadata; this greatly expands the types of media that we can handle in AzuraCast to include essentially
+  any media that Liquidsoap itself can play.
 
 ## Bug Fixes
 
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
     "azuracast/doctrine-batch-utilities": "dev-main",
     "azuracast/doctrine-entity-normalizer": "dev-main",
     "azuracast/flysystem-v2-extensions": "dev-main",
-    "azuracast/metadata-manager": "dev-add-ffmpeg-fallback",
+    "azuracast/metadata-manager": "dev-main",
     "azuracast/nowplaying": "dev-main",
     "azuracast/slim-callable-eventdispatcher": "dev-main",
     "bacon/bacon-qr-code": "^2.0",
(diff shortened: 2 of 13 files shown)


SlvrEagle23:
I'm worried about adding `application/octet-stream` because...that's basically any unidentified file. A vast majority of those _wouldn't_ be processable or playable. I know we're not inclined to do extension-based checking, but I believe for that specific MIME type, it may be necessary (and I believe is actually already done in our checker) to check by extension.

Vaalyn:
> I'm worried about adding `application/octet-stream` because...that's basically any unidentified file. A vast majority of those _wouldn't_ be processable or playable.

Good point 🤔 

> I know we're not inclined to do extension-based checking, but I believe for that specific MIME type, it may be necessary (and I believe is actually already done in our checker) to check by extension.

Having an additional "if mime type inconclusive / `application/octet-stream` then go by extension" would be suitable here I think.

The current extension check however does not have a match for the `stm` file extension in the list of extensions here: https://github.com/thephpleague/mime-type-detection/blob/main/src/GeneratedExtensionToMimeTypeMap.php

So we'd either need to add an additional lookup or extend the generated map.

Which one would you prefer to do here?

SlvrEagle23:
> Which one would you prefer to do here?

Fortunately, this League library doesn't privatize all of its methods/properties...I think we could extend that easily enough and provide a custom mapping, just so we avoid having `application/octet-stream` (and avoid doing direct extension parsing without using the library).

SlvrEagle23:
@Vaalyn Went ahead and did it, had some spare time this morning. :)

Vaalyn:
> @Vaalyn Went ahead and did it, had some spare time this morning. :)

Nice 👍 

Small question that I had in mind in regards to allowing more formats:

Should we be adding some more video formats to the list of allowed mime types to process too?
We already allow `video/mp4` to be processed (admittedly that was only added due to some m4a files wrongly being identified as that mime type) but I don't think there is anything really preventing us from allowing those files in general.

If you upload a video in mp4 it will work flawlessly atm too.
Liquidsoap just "omits" the video part of the file and pushes out the audio content.

Allowing more video formats could help people that have a lot of music videos, AMVs, etc to upload them directly without first having to convert them to audio only. (Yes they will take up a lot more disk space but that's not something we can fix for them 😉).

SlvrEagle23:
@Vaalyn I think if we're going to do that, then perhaps we should make FFProbe the default handler for metadata in general, because I think the performance advantage of getid3 vanishes when dealing with larger files, and I don't want to create the expectation that our system can gracefully handle larger files when it really can't.

PR #5321, opened 2022-04-29
Liquidsoap download link fixed

ogrenci01:
Liquidsoap download link edited.


aconitumnapellus:
@SlvrEagle23 Please backport this into stable, as stable wont work without this commit.

SC2Mitch:
We're aware sit tight

PR #5344, opened 2022-05-04
feat: add stereo tool support for stations

bjarn:
Still a work in progress, just started working on the idea.

**Proposed changes:**

- Adds native support for Stereo Tool, an industry standard for software processing used by many broadcasters
- Moves the "nrj" setting to a seperate radio input group for audio processing

- [x] Download x64 or ARM executable for Stereo Tool
- [x] Add setting to enable Stereo Tool or Liquidsoap processing
- [x] Add Stereo Tool config upload field
- [x] Add Stereo Tool license key field (validate?)
- [x] Add Stereo Tool to Liquidsoap configuration
- [x] Update NRJ method in Liquidsoap configuration
- [ ] Test :)


Vaalyn:
I have an old (outdated) plugin for AzuraCast that you can also take stuff from if you want (the UI part is outdated since we changed to Vue and removed the old form component that I used in the plugin): 
https://github.com/Vaalyn/azuracast-stereo-tool-plugin

For example in there I have the `replay_delay=1.0` parameter set for the `pipe` since this is explicitly mentioned [in the LS docs](https://www.liquidsoap.info/doc-dev/reference.html#pipe) to use when integrating steretool:
https://github.com/Vaalyn/azuracast-stereo-tool-plugin/blob/master/src/EventListener/StereoToolLiquidsoapConfig.php#L49

I'm not sure if we should have the install process automatically download stereotool.
Having an upload form in the administration like we have for SHOUTcast would probably preferable.
What do you say about this part @SlvrEagle23 ?

> Will consider if you have to upload a .sts file or paste its contents (it's just text) in a textarea. What is it you prefer?

Ideally I'd like to have a `Upload Config` button which prompts the user to upload the `.sts` file, the file uploads and then you can see / edit the contents in a textarea.

bjarn:
Didn't know about that plug-in, nice to see!

> For example in there I have the replay_delay=1.0 parameter set for the pipe since this is explicitly mentioned in the LS docs to use when integrating steretool:
https://github.com/Vaalyn/azuracast-stereo-tool-plugin/blob/master/src/EventListener/StereoToolLiquidsoapConfig.php#L49

Good one, will add.

As for downloading the binary, I was indeed thinking about a "download" button so it installs on demand. But as it also has some free features, it might be useful to less advanced setups too.

SlvrEagle23:
@bjarn Thank you for starting work on this PR!

I definitely want Stereo Tool support to be baked into AzuraCast, as we've had countless people ask about how to implement it, and without some internal support for just the binary being in place, it's been very difficult to advise folks of what to do without overriding Docker and other painful things.

One thing to note: we *cannot* ship Stereo Tool directly. It's not FOSS software, so we absolutely have to follow the same mechanism we use for SHOUTcast and instruct users to download the binary themselves, then upload it via a web interface that extracts it to a proper location and confirms its version. Probably a lot of code overlap between this and SHOUTcast.

I'm happy to throw some time into this to get it implemented. At the very least, we can get it so the binary is accessible to users via advanced configuration, and then expand from there.

bjarn:
@SlvrEagle23 Thanks a lot! Sounds good. I was waiting on your reply before continuing on this (which was a good thing to do hehe!)

Indeed a fair reason to change the installation for Stereo Tool.

Will check out your changes tomorrow and finish the PR 😁

bjarn:
CleanShot 2022-05-09 at 23 42 22@2x

Still experiencing an issue with uploading the config file of which the path is not being saved in the database _but_ is on the filesystem. It does exist.

Can't find out what exactly goes wrong here yet..

Marcioghiraldelli:
Is this project still in testing?

bjarn:
> Is this project still in testing?

Yes! Haven't had the time to give it another try yet. Will look at it this weekend probably.

Vaalyn:
@bjarn I've rebased the branch on the current Rolling-Release and added showing the currently installed version of Stereo Tool to the admin page.

I also looked at the issue you mentioned about the config file path not being saved to the DB and when I look at the DB the path is correctly saved. The problem I see however is that when reloading the `Edit Station Profile` page the `hasStereoToolConfiguration` in the Vue component is set to `false`:

image

I haven't yet looked at why this is not getting the correct value set but in the DB the path is there:

image

bjarn:
Thanks @Vaalyn! Hmm, will give it another shot asap. Last time the column stayed empty, probably an error on my side haha

Marcioghiraldelli:
I have a doubt, does this function need a separate license or would there be a free option?

Vaalyn:
> I have a doubt, does this function need a separate license or would there be a free option?

Here is the page where the available licenses for Stereo Tool are listed: https://www.thimeo.com/stereo-tool/buy/

Generally you can download the binary from their website without buying it to evaluate it first but you will have a limited feature set as far as I'm aware and legally you are obligated to buy a license from them if you want to use it for a production radio station.

For more information please contact the Thimeo Audio Technology B.V. via their forum or contact form.

Marcioghiraldelli:
@Vaalyn  is there any way I can install the stereo tools to test without the function being available on azuracast first?

Marcioghiraldelli:
@SlvrEagle23  excellent!

bjarn:
Thanks a lot @SlvrEagle23! Didn't have the time to finish it last week, but you were quick haha. Thanks!

eranwo:
@SlvrEagle23 I've upgraded  to the latest rolling release version on my testing server and so far stereo tool works like a charm! 
For the real radio people among us, this is without a doubt one of the most important upgrades made so far.
Very good!
Thanks!

PR #5462, opened 2022-06-06
hide bitrate for FLAC mount points

luuk58:
**Fixes issue:**
Mount points using FLAC show incorrect bitrate, as the encoder does not have a bitrate setting.

**Proposed changes:**
Only show bitrate when codec is not FLAC.


luuk58:
lol

PR #5463, opened 2022-06-06
hide bitrate for FLAC mount points

luuk58:
Fixes issue:
Mount points using FLAC show incorrect bitrate, as the encoder does not have a bitrate setting.

Proposed changes:
Only show bitrate when codec is not FLAC.


Marcioghiraldelli:
FLAC has a problem also related to the name of the song. When activated, the name of the song stops and stays as an offline station, but the station continues playing.

Issue #5594, opened 2022-07-13
After the update of my azuracast, my panel and the streams are no longer made https.


rolradio:
### Installation Method

Docker Installation

### AzuraCast Release Channel

Rolling Release Channel

### Current AzuraCast Version

64e2a48 (2022-07-07 2:30)

### What happened?

After the update of my azuracast, my panel and the streams are no longer made https.
A manual update gives me this in response:

[2022-07-13T04:04:36.78838+00:00] AzuraCast.DEBUG: ACME: Using directory URL: https://acme-v02.api.letsencrypt.org/directory [] []
[2022-07-13T04:04:37.523806+00:00] AzuraCast.ERROR: ACME Error: No account exists with the provided key (urn:ietf:params:acme:error:accountDoesNotExist) {"exception":"[object ] (skoerfgen\\ACMECert\\ACME_Exception(code: 0): No account exists with the provided key (urn:ietf:params:acme:error:accountDoesNotExist) at /var/azuracast/www/vendor/skoerfgen/acmecert/src /ACMEv2.php:122)"} []

skoerfgen:
I just added a pull request which fixes this issue.

If you delete/move the file `account_key.pem` (it should be in the `acme/` directory) and then try again, it should work.

rolradio:
@[skoerfgen]
Thank you very much.
That solved the problem.

PR #5600, opened 2022-07-16
Save account_key.pem after succeeded registration

skoerfgen:
Fixes issue: #5594


skoerfgen:
The problem seems to be caused when the account key is generated and then saved to disk and the following call to [register](https://github.com/skoerfgen/ACMECert#acmecertregister) fails for whatever reason. Then on the next run the issue #5594 arises, because the account key never got registered.

ItsMitchh:
Thanks for the pull! I'll approve the the checks for ya.

PR #5876, opened 2022-11-23
Add more needed permissions

scysys:
**Fixes issue:**
Nothing. Just added more permissions that i missed.

**Proposed changes:**
I also would recommend to make this interface more user friendly with a little css change.

Bildschirm­foto 2022-11-23 um 12 28 40

```
.custom-control-inline {
  display: block;
}
```


BusterNeece:
@scysys Sorry for the delay in reviewing this pull request. We can certainly look into adding new permissions, but it's a much more complex process than just changing the listed permissions in the enums, as you'd have to add those respective permission checks to every area that uses them. We'd also need to make sure that users migrating from older versions have permissions set appropriately so that if they _previously_ had access to a feature, they still have access to it with the new expanded permissions.

scysys:
Absolutly no problem Buster. Ill closed it to reorganize the fork. I need the main and stable branch on actual stand. This message 1 commit ahead and when i sync it with your main will just disturb me. I will do it in another way next time :)

Ill never worked with vue. So i used search replace to find the permission part inside azura. For what im done here it is worked for me. Im sorry when im not saw all required parts :( ^^

BusterNeece:
@scysys No problem at all; I greatly appreciate the effort to submit a pull request, and we will certainly incorporate your CSS styling change recommendation to the permissions page.

PR #5981, opened 2022-12-31
add known issue about invisible volume slider

guiltlab:
**Fixes issue:**
none

**Proposed changes:**
add known issue (volume slider appearance)


BusterNeece:
This is a fairly minor issue and doesn't belong in our "Known Issues" section. If the issue is still happening, we can just reopen the main issue. No need to add it in our Known Issues file.

PR #5991, opened 2023-01-06
link to listeners page

guiltlab:
**Fixes issue:**

**Proposed changes:**
I'm suggesting wrapping a link to the listeners reports around the Headset icon + total listeners part of the Now Playing panel in the main station profile, it's pretty handy to be able to navigate to that page from this. I added color #fff to avoid making the text blue (which is not very legible since the card header is blue by default). Not sure if this should be specified in another place for the CSS...

### DIFF
--- a/frontend/vue/components/Stations/Profile/NowPlayingPanel.vue
+++ b/frontend/vue/components/Stations/Profile/NowPlayingPanel.vue
@@ -13,14 +13,16 @@
                     class="card-subtitle text-right flex-fill my-0"
                     style="line-height: 1;"
                 >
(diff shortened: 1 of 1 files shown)


BusterNeece:
This is now implemented in the latest Rolling Release. I implemented it differently than in this PR, since as you can tell by the other parts of this component, we don't hard-code URLs into these templates, and we also have to add permission checks, as not every user who can see the profile can also see reports.

guiltlab:
I figured my PR was probably not up to your standards, I'm still learning ^^"
I'm glad you implemented the feature though, thanks!

PR #6039, opened 2023-01-22
Fixes #6037 -- Fix upload of Custom Browser Icon

### LINKED ISSUE #6037: NotWritableException at /var/azuracast/www/vendor/intervention/image/src/Intervention/Image/Image.php L150: Can't write image data to path
### Installation Method

Docker Installation

### AzuraCast Release Channel

Rolling Release Channel

### Current AzuraCast Version

xx

### What happened?

I'm trying to update the favicon and it's giving permission error.

### Relevant log output

```shell
NotWritableException at /var/azuracast/www/vendor/intervention/image/src/Intervention/Image/Image.php L150: Can't write image data to path (/var/azuracast/uploads/browser_icon/original.png
```

MarkWasley:
**Fixes issue:**
#6037

**Proposed changes:**
Correct file paths to upload custom browser icon.

### DIFF
--- a/src/Assets/BrowserIconCustomAsset.php
+++ b/src/Assets/BrowserIconCustomAsset.php
@@ -46,7 +46,7 @@ final class BrowserIconCustomAsset extends AbstractCustomAsset
         $newImage->resize(256, 256);
 
         $this->delete();
-        
+
         $destPath = $this->getPath();
         $this->ensureDirectoryExists($destPath);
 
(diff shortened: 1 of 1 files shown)


BusterNeece:
@MarkWasley I looked into this, and it wasn't necessary to make these particular changes to fix the problem. It was just an issue with the fact that `ensureDirectoryExists` was calling `dirname` on the incoming string, which made it fail to create the subdirectory. Fixing that is all that was required, no other changes needed.

PR #6141, opened 2023-03-08
Automatically fetch the latest Icecast KH Version

scysys:
Just saw the latest KH19 & KH20 commits. It seems you would stay up to date with Icecast KH. So i think this will help you to save some time. If you want , this would be a way you can make sure your script is downloading automatically the latest version.

**Fixes issue:**

**Proposed changes:**


BusterNeece:
So, there's actually a reason why we *don't* do this: because our Docker image builder aggressively caches build steps, and if we did it this way, there would be no change between versions in the build files, so the build process would have no way of knowing it needs to actually update the Icecast version. We'd end up having to change the file in some way _anyway_, so we might as well just explicitly specify a version to ensure predictable results.

PR #6246, opened 2023-04-30
Replace deprecated LS operators & fix a warning

Vaalyn:
**Fixes issue:**
x

**Proposed changes:**
I've fixed 2 deprecations and a warning about an unused variable I've seen while debugging the startup issue in LS 2.2.0.

Deprecations / Replacements:
```
setenv => environment.set
add_protocol => protocol.add
```

My IDE also automatically removed unnecessary spaces from the LS config blocks.


BusterNeece:
@Vaalyn Aren't those spaces needed for indented heredoc blocks like that? My IDE seems to think they are :o

Vaalyn:
Nope, they are not needed.
Neither PHP nor LS breaks after removing them.

PR #6249, opened 2023-05-01
Implement Liquidsoap 2.2.x Rolling Release

BusterNeece:
A rolling PR to let us address any lingering issues with 2.2.x before implementing it into our Rolling Release.

<a href="https://gitpod.io/#https://github.com/AzuraCast/AzuraCast/pull/6249"></a>


Moonbase59:
Upstream bugs https://github.com/savonet/liquidsoap/issues/3039 and https://github.com/savonet/liquidsoap/issues/3040 have been fixed according to @toots, so this might be tried again, if there is time for it.

BusterNeece:
Updated the Rolling Release on this build; I think it merits a little more testing before pushing back to Rolling Release, lest we end up doing the revert dance again.

Moonbase59:
> I think it merits a little more testing

Happy to test. Would there be an easy way to switch a docker instance between https://github.com/AzuraCast/AzuraCast/tree/main and https://github.com/AzuraCast/AzuraCast/tree/feature/liquidsoap-2.2.x? Preferably without needing to build and/or spin up an extra VM?

Issue #6332, opened 2023-06-09
Could we move Liquidsoap to next commit to fix blank.eat()?


Moonbase59:
### Installation Method

Docker Installation

### AzuraCast Release Channel

Rolling Release Channel

### What happened?

@BusterNeece Can we dare to move Liquidsoap 2.2 to ~~commit #50cc747~~ the next commit once the build finishes, so we can test `blank.eat()`? It is supposed to fix https://github.com/savonet/liquidsoap/issues/3132 which hangs my station when songs have silence at the end.

This _is_ a bumpy road, but thanks to all participating, we’re really getting somewhere…

Moonbase59:
@Vaalyn We’d need commit 50cc747 for the fixed blank.eat() but their build failed. Using 39717ae might be too early a version.

Moonbase59:
Confirm `blank.eat()` works as intended in AzuraCast 2e0c89d with LS 2.2.0+git@50cc747f6. It _does_ use quite a bit CPU when "eating" longer silence parts (went to ~80% shortly), but quickly goes down again, and playout is fine.

Anyway, a typical radio would mostly use `blank.skip()` instead, as shown in https://github.com/AzuraCast/AzuraCast/discussions/6334, which also works fine.

Moonbase59:
Additional feedback re LS 2.2.0+git@50cc747f6:

- Runs stable at 12-18% CPU since about 17 hours now, with HLS enabled + MP3 + AAC stream.
- "Use ReplayGain" in station disabled (using pre-tagged files) since that still eats too much CPU. I think @toots is working on that.
- Now enabled additional "master_me" processing and will leave it running for a day or so. Brings CPU load up to ~28–32%, which is to be expected. 15-min load avg around 0.9–1.0, with 1-min avg sometimes going up to 1.6 when doing tasks like CheckMediaTask. Seems okay so far.

toots:
See: https://github.com/AzuraCast/AzuraCast/issues/6327#issuecomment-1585680594 for additional optimization/regression fix of replaygain.

PR #6336, opened 2023-06-09
Update Liquidsoap to RR 2.2.0 #39717ae

Vaalyn:
**Fixes issue:**
Fixes https://github.com/AzuraCast/AzuraCast/issues/6332

**Proposed changes:**
Updates LS to the newest RR commit that includes some further fixes

### DIFF
--- a/util/docker/stations/setup/liquidsoap.sh
+++ b/util/docker/stations/setup/liquidsoap.sh
@@ -20,7 +20,7 @@ if [[ "$(uname -m)" = "aarch64" ]]; then
 fi
 
 # wget -O /tmp/liquidsoap.deb "https://github.com/savonet/liquidsoap/releases/download/v2.1.4/liquidsoap_2.1.4-ubuntu-jammy-1_${ARCHITECTURE}.deb"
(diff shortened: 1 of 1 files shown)


Moonbase59:
@Vaalyn We’d need commit 50cc747 for the fixed `blank.eat()` but their build failed. Using 39717ae might be too early a version.

Vaalyn:
@Moonbase59 thank you checking. It seems there was an error in the Liquidsoap build pipeline on their repository: https://github.com/savonet/liquidsoap/actions/runs/5216092480

Which resulted in it not being pushed to the [release assets repository for LS](https://github.com/savonet/liquidsoap-release-assets/releases/tag/rolling-release-v2.2.x)

Can you notify them about this so they can re-run that pipeline?

Moonbase59:
Tried to notify @toots already, hopefully he’ll see it.

toots:
Thanks for the notification. The build is finishing now.

Also worth mentioning, this build on debian will now have ffmpeg with libfdk_aac enabled in it. There's been some feedback that `%ffmpeg` with `libmp3lame` and `libfdk_aac` is much less CPU intensive. We're not sure yet why but worth considering!

toots:
(I meant docker debian above, which is the default production docker, as opposed to the alpine docker -- for which I have no idea if ffmpeg supports libfdk_aac)

Moonbase59:
@toots Thanks for fixing! Checking out ffmpeg with `libfdk_aac` is surely worth a try.
@Vaalyn The build went though now, could we integrate?

BusterNeece:
Committed an update directly to the commit requested.

PR #6387, opened 2023-07-02
Added dvh for public pages

zcraber:
**Proposed changes:**
Adding dynamic viewport height for public pages as web browsers have started [supporting](https://caniuse.com/?search=dvh) it. This will make the footer on public pages visible by default on mobile devices.

azura

In case the browser is not supported `height: 100vh;` will be used as a fallback. 

<a href="https://gitpod.io/#https://github.com/AzuraCast/AzuraCast/pull/6387"></a>

### DIFF
--- a/frontend/scss/azuracast/pages/_public.scss
+++ b/frontend/scss/azuracast/pages/_public.scss
@@ -3,6 +3,7 @@
 body.page-minimal {
   .public-page {
     height: 100vh;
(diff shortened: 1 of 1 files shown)


BusterNeece:
Thank you for the information! We just did a full update of our styling to use Bootstrap 5, and we incorporated your suggestion into the new styles, so the public page is using the proper height.

PR #6394, opened 2023-07-05
Update Frontend: +Vue 3, -Vue Compat, +Bootstrap 5 -Materialize -BootstrapVue +Oruga

BusterNeece:
This set of changes brings us fully into the present-day versions of our frontend software, including both Bootstrap 5 and Vue 3.

Because of BootstrapVue Next's unfinished and non-production-ready status, any components that depended on it have either been replaced by direct implementations using the Bootstrap HTML or the framework agnostic library Oruga.

All major components have been tested, and while they may look a little different, they still function just as well for our purposes.

This needs some significant testing before merging, however.

<a href="https://gitpod.io/#https://github.com/AzuraCast/AzuraCast/pull/6394"></a>


Vaalyn:
@BusterNeece while playing around and doing some adjustments on the new styling I noticed an unpleasent effect when switching tabs on a `nav-tab` element that we should fix. You can see the content of the tab outside of the card while switching tabs, see attached video.

https://github.com/AzuraCast/AzuraCast/assets/13745863/b7457957-4bad-4848-ac96-21cf0b5061ec

BusterNeece:
@Vaalyn Fixed!

Vaalyn:
👍 thanks, I've also created a draft PR were I'm collecting some tweaks / adjustments / fixes for this update.

When I'm done with that I'll convert the draft PR into a real PR but feel free to check it out already.

PR #6412, opened 2023-07-10
Fix RAM usage calculation error

Vaalyn:
**Fixes issue:**
X

**Proposed changes:**
I've looked into why our RAM usage panel and `htop` show so different values (see [this thread](https://github.com/AzuraCast/AzuraCast/discussions/6406#discussioncomment-6397734)) and compared our calculation to the C code of htop and implemented 3 fixes:

- Consider `buffers`, `sReclaimable` & `shmem` in calculation for RAM usage
- Set number base from `1000` to `1024` in `Quota.php`
- Increased decimals in readable values to 2 places to reduce rounding differences

I changed the number base to 1024 because all places that I looked at in the CLI (namely `df`, `htop`, `du`) showed different values than we do and after changing the base they started agreeing.

I didn't notice any place in de UI that looked wrong now so I think we were already expecting those to be base `1024` although we really calculated based on `1000`.

<a href="https://gitpod.io/#https://github.com/AzuraCast/AzuraCast/pull/6412"></a>

### DIFF
--- a/src/Controller/Api/Admin/ServerStatsAction.php
+++ b/src/Controller/Api/Admin/ServerStatsAction.php
@@ -150,14 +150,20 @@ final class ServerStatsAction implements SingleActionInterface
                 'bytes' => [
                     'total' => $memoryStats->memTotal,
                     'free' => $memoryStats->memFree,
-                    'cached' => $memoryStats->cached,
+                    'buffers' => $memoryStats->buffers,
+                    'cached' => $memoryStats->getCachedMemory(),
+                    'sReclaimable' => $memoryStats->sReclaimable,
+                    'shmem' => $memoryStats->shmem,
                     'used' => $memoryStats->getUsedMemory(),
                 ],
                 'readable' => [
-                    'total' => Quota::getReadableSize($memoryStats->memTotal),
-                    'free' => Quota::getReadableSize($memoryStats->memFree),
-                    'cached' => Quota::getReadableSize($memoryStats->cached),
-                    'used' => Quota::getReadableSize($memoryStats->getUsedMemory()),
+                    'total' => Quota::getReadableSize($memoryStats->memTotal, 2),
+                    'free' => Quota::getReadableSize($memoryStats->memFree, 2),
+                    'buffers' => Quota::getReadableSize($memoryStats->buffers, 2),
+                    'cached' => Quota::getReadableSize($memoryStats->getCachedMemory(), 2),
+                    'sReclaimable' => Quota::getReadableSize($memoryStats->sReclaimable, 2),
+                    'shmem' => Quota::getReadableSize($memoryStats->shmem, 2),
+                    'used' => Quota::getReadableSize($memoryStats->getUsedMemory(), 2),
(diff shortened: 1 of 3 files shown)


Moonbase59:
Good find about Quota using 1000’s.

I wouldn’t go as far as include reclaimable kernel buffers in the calculation, or shmem (that’s included in Cached, anyway).

Take `free` as an example, which is probably the most basic code for this type of calculation. You’ll see that

```
used = total - buffers - cache -free
```

should suffice (and be exactly what `free` is showing.

`top` should (hopefully) come up with the same values, and `htop` is, as admitted by its author, "guessing" in places.

I’ve spent some hours now checking source code of `top`, `free` and `htop` (in parts), and everyone seems to agree that above calculation is what should be used for "used".

Memory in Linux is a rather complicated issue, and has some pitfalls and traps.

Vaalyn:
I tried that but then the calculation was showing around 300-400mb too low as used.

The only way I've got this to match consistently was by doing it like I've done here.

Moonbase59:
Match with what? top? htop? free?

Vaalyn:
> Match with what? top? htop? free?

htop

Moonbase59:
The least exact of the three … I’d go for above. Better exact than "matching" … ;-)

Vaalyn:
I'd prefer showing the same value as htop since in most cases I've encountered, people are looking at htop and not top or free when reporting cpu / memory usage to us as that's better human readable.

Moonbase59:
I prefer "correct", and it has paid in the decades I’ve been programming.

There are a lot of intricacies in Linux memory management. Just think of … someone doing a ftruncate in a buffer, a zombie process not freeing its shared memory, and the like. And there was a reason for including shmem in Cached. And so on…

Vaalyn:
Well, `free` and `top` show the shared memory seperately, `htop` includes it and doesn't show it.

If someone has a problem with free / top not agreeing with htop / our panel they have a way to quickly see why, namely using this shared memory value to calculate it.

If such a zombie process occurs then htop would also baloon its displayed value in size.

I frankly don't see the general issue with how it is calculated now apart from a small percentage of users who might be using top / free instead of htop having to do some 1 + 1 instead of htop users having to do 1 - 1 and htop users don't even see that value separately so for those it would be much harder to do.

Moonbase59:
Actually, htop shows it in the color bars (inexact as hell, of course): green=used, blue=buffers, yellow=cache.

Vaalyn:
image

image

Those values are matching, when I change it to not include `shmem` & `sReclaimable` it differs by around 300mb.

In `htop` you have no number value shown for the shared memory so there is no way for `htop` users to calculate why we would show 3.69 instead of 3.99.

On the other hand, `top` and `free` do show you that number, so those user can calculate it.

I don't want to make this harder for people using `htop` which is much more common imho.

Correctness doesn't really matter that much when both are correct in their own context, especially if one part of the users can't get part of the equation since the program doesn't show that part.

I won't be changing this just for the sake of absolute technical correctness.

PR #6648, opened 2023-10-06
feat: Added Elestio as fully managed deployment option

kaiwalyakoparkar:
**Fixes issue:**
In support with the suggestion made here https://features.azuracast.com/suggestions/499616/adding-elestio-as-fully-managed-deployment-option

**Proposed changes:**
Elestio has been providing options of fully deploying and managing AzureCast application [as shown here](https://elest.io/open-source/azuracast). I think it would be a great idea if we can add it to official readme/documentation here. I have also share about the support we provide to the project in the suggestions above. Do let me know :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,7 @@ We would like to thank the following organizations for their support of AzuraCas
 - [BrowserStack](https://www.browserstack.com/) for giving us tools to help test our software's accessibility and
   features across multiple browsers and operating systems.
 - [Depot](https://depot.dev/?utm_source=AzuraCast) for powering our Docker image builds.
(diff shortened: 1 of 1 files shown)


BusterNeece:
Thank you for reaching out! I've sent you an e-mail to follow up on the plan Elestio offers to help sustain open-source projects.

In the case of this PR, I'm closing it for now, as it's currently listed in the "thank-yous" section, which is specifically where we thank the companies who provide us with complimentary services that they would normally charge for. If we were to promote Elestio, it would be on our homepage and documentation, and we'd want to make sure we had something arranged in place before doing so.

PR #6808, opened 2023-12-20
Add missing HLS default API interfaces.

PeWe79:
**Fixes issue:**

Fixes missing API interfaces HLS default.
**Proposed changes:**

Add hls_is_default on ApiInterfaces.


BusterNeece:
Thank you for the PR! We do this automatically though via internal PHP and NPM scripts. I just hadn't regenerated the interfaces quite yet. That's done now in the latest Rolling Release.

PR #6855, opened 2024-01-09
Update default.po

angrystar170:
**Fixes issue:**

**Proposed changes:**


BusterNeece:
@angrystar170 Hello! Thank you for your contribution! Translations should be submitted at https://translate.azuracast.com/ instead of directly modifying our translation files. Thank you!

Issue #6956, opened 2024-02-19
Let's Encrypt certificate expired but not updated


wunschradio:
### Installation Method

Docker Installation

### AzuraCast Release Channel

Stable Release Channel

### Current AzuraCast Version

v0.19.4 Stable • Docker • PHP 8.2

### What happened?

Hi. The Let's Encrypt Certificate expired yesterday. The acme cron (checking every 3 hours) wrote that it is not expired. Therefor no new update has been started. I had to delete the acme.key in /var/lib/docker/volumes/azuracast_acme/_date to get a new certificate.

### Relevant log output

```shell
AzuraCast.DEBUG: ACME: Using directory URL: https://acme-v02.api.letsencrypt.org/directory [] []
AzuraCast.ERROR: ACME Error: Certificate does not need renewal. {"exception":"[object] (RuntimeException(code: 0): Certificate does not need renewal. at /var/azuracast/www/src/Service/Acme.php:136)"} []
```

wunschradio:
Thanks!

Issue #6918, opened 2024-02-05
letsencrypt certificate refresh without nginx reload


wendelinux:
### Installation Method

Docker Installation

### AzuraCast Release Channel

Stable Channel

### Current AzuraCast Version

0.19.3

### What happened?

after a while i notice certificate invalid notifications in my browser. When asking for a new certificate azuracasts says that there are no certificates due for renewal.
when i reboot the system the certificate is valid without dooing a renewal.
it looks to me that when the letsencrypt certificate gets updated, nginx is not beeing reloaded. Without reload it'll keep using the old and expired certificate

BusterNeece:
Fixed in the latest version of the Rolling Release. Thanks, @skoerfgen!

PR #6991, opened 2024-03-04
Fix for issues #6918 and #6956

skoerfgen:
**Fixes issue:**
#6918 and #6956

**Proposed changes:**
- call to `reloadServices` was missing after renewal

In case of renewal (where the symlinks are already present) the `checkLinks` method did not call `reloadServices`. So the certificate got successfully renewed but the webserver reload was missing.

I rearranged the code to make sure `reloadServices` is called on renewal and when the certificate is still valid but the symlinks are missing.

This was introduced by commit 5acf5d7f2ae5e50ecbd432753a895dddfc58e5e0


BusterNeece:
@skoerfgen Ah, thank you for your contribution! I'm glad the problem reported there turned out to be a simple situation.

skoerfgen:
@BusterNeece I'm glad I could help!

PR #7001, opened 2024-03-09
Now playing position

jpcastberg:
**Fixes issue:**

**Proposed changes:**
I'd like to add position state to the public radio page. This will allow native media interfaces (like you'd see on the android or iphone lock screen) to display song duration and time elapsed + a progress bar.

See MDN: https://developer.mozilla.org/en-US/docs/Web/API/MediaSession/setPositionState

Tested on iphone 13 running ios 17.4 (note, i had to install the public page as a webapp before the media interface populated w/ the metadata).

One unexpected bug I had to fix - attempting to seek using the progress bar broke the stream. I fixed this by adding a seeking handler that resets the progress bar to the last known position. Last known position is usually a little behind where the stream actually is, but the next server side event heals that.

Here's a video of the result which includes the "seek" behavior.
https://github.com/AzuraCast/AzuraCast/assets/37011196/d2d47342-18bf-48d3-a5f6-724df7ca2657

Apologies if i needed to open an issue or something first - but this seemed like an easy freebie :)


frontend/src/functions/useNowPlaying.ts
```diff
@@ -46,11 +46,24 @@ export default function useNowPlaying(props) {
                     {src: np_new.now_playing.song.art}
                 ]
             });
+
+            setPositionState(np_new.now_playing.duration, np_new.now_playing.elapsed);
+
+            navigator.mediaSession.setActionHandler("seekto", function () {
+                setPositionState(np_new.now_playing.duration, np_new.now_playing.elapsed);
+            });
```
jpcastberg, line 54: From pr description:
> One unexpected bug I had to fix - attempting to seek using the progress bar broke the stream. I fixed this by adding a seeking handler that resets the progress bar to the last known position. Last known position is usually a little behind where the stream actually is, but the next server side event heals that.

BusterNeece:
Thank you for your contribution! Looks good to me.

PR #7044, opened 2024-04-05
#6252 -- Update to latest LS RR for protocol resolving fixes

RM-FM:
**Fixes issue:**
Fixes an issue with missing metadata id using custom protocols and `metadata.decoder.add`. With this, autocue will respect cue/cross/fade values set manually in Azuracast.

### DIFF
--- a/util/docker/stations/setup/liquidsoap.sh
+++ b/util/docker/stations/setup/liquidsoap.sh
@@ -19,7 +19,7 @@ if [[ "$(uname -m)" = "aarch64" ]]; then
     ARCHITECTURE=arm64
 fi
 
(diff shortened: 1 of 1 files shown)


RM-FM:
@BusterNeece There was an issue related to missing metadata in `metadata.decoder.add` if a custom protocol was user. As Azuracast uses its own `media` protocol, `autocue ` behaved unpretictable as soon as one maintained manual cue/fade values in Azuracast. With this fix, Romain made sure that all metadata is available LS internally, even with custom protocols. As a result of this fix, `autocue` is able to reliably recognize existing cues/fades and pass them through, instead of trying to calculate them again (creating inconsitent data) and finally messing up the transitions.

PR #7129, opened 2024-05-20
Chown /var/azuracast folder to match updated puid/guid

jpcastberg:
**Fixes issue:**
n/a

**Proposed changes:**
If the azuracast user and group ids are changed as the result of the AZURACAST_(P|G)UID environment variables, the azuracast user no longer owns the `/var/azuracast` directory, causing permissions issues. Here I have recursively updated ownership of the directory after the new ids have been set.

Open issue: It's possible the user may have bind mounted their own directories into `/var/azuracast`, which would change ownership of their files if they did not make the mounts read-only - this could be mitigated by not recursively chowning `/var/azuracast`, and recursively chowning children of that directory that are more likely to be managed by azuracast only (stations, backups, storage, etc)

### DIFF
--- a/util/docker/web/startup_scripts/00_setup_user.sh
+++ b/util/docker/web/startup_scripts/00_setup_user.sh
@@ -11,3 +11,5 @@ usermod -o -u "$PUID" azuracast
 
 echo "Docker 'azuracast' User UID: $(id -u azuracast)"
 echo "Docker 'azuracast' User GID: $(id -g azuracast)"
+
+chown -R azuracast:azuracast /var/azuracast


BusterNeece:
@jpcastberg Well, that's kind of two different ways of solving the same problem, right?

The reason we have the PGID/PUID env vars is so that you can override the UID/GID inside Docker so it _matches_ the UID/GID that owns the bind-mounted directories in the host machine. That way we're not just trying to steamroll and chown everything in that directory, which would not only be a huge performance/IO hit for lots of stations but would fail in lots of edge cases, such as read-only bind mounts.

If you're using custom bind mounts, just make sure that the user specified in PUID/PGID owns the directory on the host machine.

jpcastberg:
@BusterNeece  I found the issue - I am using bind mounts for subdirectories of `/var/azuracast` (db_data, acme, shoutcast2_install, etc...).

When using a bind mount, and the id of the azuracast user is changed to `$AZURACAST_PUID`, the owner of `/var/azuracast` does not change - it stays owned by user 1000 and group 1000, no longer associated with the azuracast user. The permissions on that directory are `drwx------`, effectively banning the azuracast user from accessing that directory, causing the problems that prompted me to open this pr.

Using a docker volume, when the azuracast user's id is updated, `/var/azuracast` stays owned by azuracast, now with the changed id. No access problems at all.

I like having bind mounts as sometimes it can be helpful to dig around in the container's files to diagnose issues, but I'll see if I can make Docker volumes play nice with my setup. Appreciate all of the work you do!

Issue #7185, opened 2024-06-08
Have Turbo mode env option for radio streams quick start, to reduce startup lag and cpu spike


t0mtaylor:
### Installation Method

Docker Installation

### AzuraCast Release Channel

Rolling Release Channel

### Current AzuraCast Version

Affects Latest and Stable

### What happened?

Have Turbo mode env option for radio streams quick start, to reduce startup lag and cpu spike

During startup, it is too slow to start streaming the station after you update or simply restart the docker container, with a massive CPU spike

Even on a system with 12 cores and 64gb of ram, the massive resource spike always slows starting up and streaming of all stations - even just one station!

Looking into the startup logs, Azuracast is trying to scan playlists and files, which seems unnecessary at startup!

I propose a "Turbo" mode within the env file, which would force Azuracast to "Quick Start" the radio streams as a priority without the massive CPU Spike with the other lower priority services and actions - bit like a feature flag

The scheduled tasks (and any others not a priority to start the stream) should also start 5 minutes after the station starts, or the user has the option to manually trigger the jobs via the GUI if required, although for Turbo mode this is not essential

BusterNeece:
@t0mtaylor Much of the startup stuff that we do isn't really something you can skip on. For example, the initial portion is flushing caches, clearing temp files, ensuring the database is on the latest version and migrated, etc...and those don't take too terrifically long but are very important.

I believe the slowdown you're referring to specifically comes from the initial CPU burden of either analyzing files for ReplayGain or for AutoCue. Both of those will, if there isn't any metadata already written to the file, basically analyze any file that _could_, in theory, play right at that moment; that includes at least one file from every playlist you have and the first file that's cued up for playback by the AutoDJ.

In that regard, there's definitely room for improvement, and it's a very high priority for us right now to do that. We have to make some "under the hood" changes to facilitate it, but we intend to add the ability for you to set a "save metadata" mode, enabled by default if you enable either ReplayGain or AutoCue, where the special metadata used by those processes is automatically written to the file (and/or to our metadata database) as it's uploaded (or during routine reprocessing), which would then be referenced by those tools upon startup, all but removing the initial CPU load and shifting it to a much more gradual process that happens only when media is uploaded or reprocessed.

t0mtaylor:
Thank you @BusterNeece 

[...]

Issue #7175, opened 2024-06-04
AzuraCast 0.20.1 Update Issues: ReplayGain and AutoCue Problems


BornKillaz:
### Installation Method

Docker Installation

### AzuraCast Release Channel

Rolling Release Channel

### Current AzuraCast Version

AzuraCast • v0.20.1 Stable • Docker • PHP 8.3

### What happened?

Ok, so today I update Azuracast to versions 0.20.1... What a disaster!

I come across similar issues I encountered on the previous update, my station haven't start, it just keep playing the fallback mp3 file. 

After a few restarts, including Liquidsoap reload, full station restart and docker restart I was able to put the station online again but only by disabling ReplayGain in the Auto DJ tab.

Then I went to activate autocue, added my settings (bellow) to liquidsoap, disabled ReplayGain and reload / restart / did everything again, with no luck the station kept offline.

```
settings.autocue.cue_file.noclip := true 
settings.autocue.cue_file.write_tags := true
settings.autocue.cue_file.target := -17.0
```

So the only way I could get my station stream back again, was to disable ReplayGain, then reload and it worked, at least it produced some sound from playlists....

So this latest version it a totally disaster for me, at first caused by ReplayGain as far as I could figure it. Now I have the station running without ReplayGain, nor autocue. And on top of that,  my playlist is constantly failing and falling back to the default playlist, and suddenly back to the correct playlist... What a mess...

### Relevant log output

```shell
https://pastebin.com/VfyQfCXR
```

BornKillaz:
On top of the described above, even when I have Smart Mode active and autocue disabled, Most of the times I get no transition at all:
2024/06/04 23:18:33 [cross.smart:3] No transition: using default.

I'm going back to previous version v20.0. Hopefully I can download with an existent backup.

BusterNeece:
This isn't a very good GitHub issue report. It's not very helpful to us to just say "this update was a disaster!" and other "the old version was better" type sentiments.

If you enabled AutoCue for the first time in 0.20.1 and that caused problems, then it's possible that it was the _enabling of AutoCue_ that caused the problems, not the actual 0.20.1 update itself, which not only doesn't require that you use AutoCue but doesn't enable it by default.

[...]

PR #7189, opened 2024-06-09
Allow disabling generation of LS playlist definitions if not needed

Vaalyn:
**Fixes issue:**
This feature could alleviate the CPU load problems from the initial analyzing of files for AutoCue, which could fix some of the load issues from #7175, #7178 & #7185

**Proposed changes:**
This PR adds a new flag to the Station Backend Configuration that let's a user control if the Liquidsoap config writer should always generate `playlist_xyz` definitions or if it should omit generating those that do not explicitly need to exists in LS code.

This will only target playlists that are song based, not advanced type, do not have any advanced options active and only on stations without manual autodj active.

This ensures that only if there is a reason for the playlist to exists in the LS code will the `playlist_xyz` definition be generated. For example due to having to manually schedule it or because it's playback relies on specific LS operators being applied for it (like switches and stuff).

Since I'd categorize this as an optimization feature that will mostly only really be noticeable by bigger stations, stations with long files, with AutoCue enabled or a combination of those I've made it set to true by default (so always generate everything). But this should make it possible if needed to reduce the amount resources consumed by LS on startup due to having potentially a lot less playlists / files to analyze on statup and to keep track of internally.


Vaalyn:
@BusterNeece 😄 naming, the hardest thing in dev. No worries, I had my issues finding something half way decent anyways and still found it clunky.

t0mtaylor:
Looks good @Vaalyn @BusterNeece - trying this now, i'll let it run the Latest RR for a couple of hours and try restarting to see if its any quicker and less cpu crazy!

(Rolling Release #[117f4cb](https://github.com/AzuraCast/AzuraCast/commit/117f4cb92bbe9d4dcadde3a9bca01efbd7344a19) (2024-06-10 9:21) • Docker • PHP 8.3)

PR #7221, opened 2024-06-21
feat:hide play button and change radio logo

OthyTenk:
What changes:
- It is updated with a new DC logo, an old one on the player widget, and an admin dashboard.
- It hid the play button. Then it did auto-play the stream.

**Fixes issue:**

**Proposed changes:**


BusterNeece:
This isn't the first pull request we've accidentally received from BoostCast.

We're happy to receive pull requests that improve the functionality or stability of the application, but we don't want PRs that incorporate things like what BoostCast is doing (branding changes, naming changes, and removing credit and donation attributions).

PR #7232, opened 2024-06-23
If not within Docker, load the azuracast.env file

Yahav:
When running inside Docker, it takes care of injecting the azuracast.env file to the local ENV. If AzuraCast runs outside of Docker it now loads the environment directly from the azuracast.env file.

**Fixes issue:**
This issue wasn't reported yet

**Proposed changes:**

- Load azuracast.env file when not running within Docker.

### DIFF
--- a/backend/src/Environment.php
+++ b/backend/src/Environment.php
@@ -6,6 +6,7 @@ namespace App;
 
 use App\Enums\ApplicationEnvironment;
 use App\Enums\ReleaseChannel;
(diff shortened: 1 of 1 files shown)


Yahav:
ps. i do feel that importing AzuraCastEnvFile App\Installer\EnvFiles\ isn't the best practice but i didn't want to modify the namespace tree by moving the EnvFIles dir to main src dir.

PR #7293, opened 2024-07-04
POC: Feature/max bitrate

Yahav:
This adds the ability for administrators to set a maximum bitrate for a station.
This will limit the bitrates of:
- Mount points
- HLS Streams
- Remote Relays
- Live Streams Recordings

It includes a backend validation as well as a mechanism in which when downgrading a station bitrate limit it also sets any of the above that is higher than the limit to the new limit.

It also utilizes Icecast-KH limit-rate functionality to limit live streaming bitrate as well, in case one was to broadcast at a higher bitrate the stream will be "skippy"/"glichy".

Requested at: https://features.azuracast.com/suggestions/113370/station-bitrate-limits

**TODO:** 

- [X] For some reason it fetched my personal changes to `frontend/entities/ApiInterfaces.ts`, needs to replace Caster.fm with the original Azuracast strings.
- [X] Same goes for changes already merged in the past, no idea how to get rid of those commits in here.
- [ ] Should probably add a notice at the "Broadcasting Service" pane as well as the "Streamer/DJ Accounts" > "Connection Information" pane letting the user know his maximum bitrate for live streams.
- [ ] As with any project there are standard and practices that i probably didn't met, this will require a thorough review.
- [X] I removed the custom bitrate option entirely as i don't see a real value for it, but if you could think of a way to implement if while retaining the limit, great.
- [ ] Not sure how to deal with the translations, not sure how do i update the translation files with the new phrases.


Vaalyn:
❤️  thank you for working on this feature.

I know this is still WIP but I've taken a look at the code already and wanted to bring something up for discussion:

The new `max_bitrate` setting is initialized with 128 as default & as fallback if none is set. I think it would be better to have the default be empty / null in order to still allow stations with arbitrary bitrates. Otherwise users will have to always configure a bitrate limit for stations even if they don't need / want that. It would also currently limit all existing stations to 128 after the migration which could most likely cause issues for existing stations.

Apart from that you could probably simplify the code in the `lowerXyzBitrate` methods a bit by using the sub-entities from the station entity instead of fetching a new instance, for example: 
```php
public function lowerHlsBitrate(Station $station): void
{
    foreach ($station->getHlsStreams() as $hlsStream) {
        if ($hlsStream->getBitrate() > $station->getMaxBitrate()) {
            $hlsStream->setBitrate($station->getMaxBitrate());
            $this->em->persist($hlsStream);
        }
    }

    $this->em->flush();
}
```

Vaalyn:
Just saw that shoutcast also seems to support setting a max bitrate in their config. You might want to add it there too:
https://sc-mirror.shoutca.st/docs/DNAS_Server.html#Stream_Configuration

Yahav:
>Apart from that you could probably simplify the code in the lowerXyzBitrate methods a bit by using the sub-entities from the station entity instead of fetching a new instance, for example:

Ohhh that's nice... will take care of it tomorrow.

>Just saw that shoutcast also seems to support setting a max bitrate in their config. You might want to add it there too

Will take care of it tomorrow, however i don't have a license for Shoutcast so i won't be able to test it.


>The new max_bitrate setting is initialized with 128 as default & as fallback if none is set. I think it would be better to have the default be empty / null in order to still allow stations with arbitrary bitrates. Otherwise users will have to always configure a bitrate limit for stations even if they don't need / want that. It would also currently limit all existing stations to 128 after the migration which could most likely cause issues for existing stations.

Hmm... so, set it to null and at the enforcements of the bitrate limit check if not null first? 
How would we handle previously entered custom bitrates? maybe there's a way to keep the custom bitrate option but limit it as well to the max station limit.

Vaalyn:
> How would we handle previously entered custom bitrates? maybe there's a way to keep the custom bitrate option but limit it as well to the max station limit.

Maybe we could use a specific validator on the property to enforce a max on there even for custom values.
The validation the frontend uses is built on [Vuelidate](https://vuelidate-next.netlify.app/) which should be able to handle such a case via the `maxValue` built-in validator.

Yahav:
@Vaalyn Done.
I addressed the backwards computability by considering 0 as unlimited and setting it as default.
This does need some further testing though.

Yahav:
Also worth mentioning that custom bitrates may break liquidsoap as it expects a standard bitrate.
```
2024/07/05 09:41:26 [lang:3] API feedback - Response (200): true
At /var/azuracast/stations/azuratest_radio/config/liquidsoap.liq, line 253, char 48-50:
    %mp3(samplerate=44100, stereo=true, bitrate=66),

Error 8: Invalid bitrate value
At /var/azuracast/stations/azuratest_radio/config/liquidsoap.liq, line 253, char 48-50:
    %mp3(samplerate=44100, stereo=true, bitrate=66),
```

Yahav:
Another note: 
This breaks the edit modal at /admin/stations.

Yahav:
Closing this one in favor of the newer one

PR #7335, opened 2024-07-19
"Create New Key" -> "Create Another Key"

mdrxy:
**Proposed changes:**
"Create New Key" doesn't indicate that an API key was just generated -- instead, it can be seen as a second button to "confirm" the creation of the API key, when it is in fact already created, and pressing the button again would generate a second key unnecessarily. May or may not be inspired by a personal experience/hiccup.


BusterNeece:
@mdrxy By the way, just so you know since you're doing a bunch of these PRs...you don't need to (and we don't necessarily benefit from) updating the translation strings any time you change a string in the app. The strings are automatically pulled for the current version of the app and synced with CrowdIn as part of the build process, which would overwrite your changes anyway.

frontend/components/Account/ApiKeyModal.vue
```diff
@@ -54,7 +54,7 @@
                     :class="(v$.$invalid) ? 'btn-danger' : 'btn-primary'"
                     @click="doSubmit"
                 >
-                    {{ $gettext('Create New Key') }}
```
BusterNeece, line 57: I don't want to solve this problem this way. If the "create new key" button is showing up on the "new key" page, that's an error and only the "close" button should be shown there. The "Create new key" button would force a resubmit of the previous data, which is certainly not what users want, regardless of what it's named.

mdrxy:
@BusterNeece good to know, thank you! Are there any places outside of the actual `.vue` files that language should be added (that aren't the translation files?

BusterNeece:
No need, the "extract strings" functions work on both the PHP and Vue files automatically.

frontend/components/Account/ApiKeyModal.vue
mdrxy: Screenshot 2024-07-19 at 03 35 34

This is the screen I'm referring to - pressing "create new" generates a second, unique key.

BusterNeece:
I think this should actually be three separate PRs, one for each thing being changed (the API key modal, the RadioReg typos/descriptions, and the vuelidate changes) because I don't necessarily think we should merge them all and I'd like to review them separately.

frontend/components/Account/ApiKeyModal.vue
BusterNeece: I changed/fixed this in the latest Rolling Release, so no PR is necessary for this feature.

PR #7362, opened 2024-08-02
Update BaseUrlCheck to include clickable link in error message

mdrxy:
**Proposed changes:**
Make the link in the Base URL notification clickable.


BusterNeece:
We don't inject HTML into notifications specifically to avoid having to worry about escaping incoming content. In my local tests, this never worked properly as a result of this. Please make sure to test your PRs locally first before submitting them or taking them out of "Draft" status.

PR #7372, opened 2024-08-06
Fix issue #7367 - Add overflow ellipsis

### LINKED ISSUE #7367: Left sidebar width issue in greek locale
### Installation Method

Docker Installation

### AzuraCast Release Channel

Rolling Release Channel

### Current AzuraCast Version

#d7070dd (2024-07-21 15:48) 

### What happened?

Verifying the greek translation.
When translated into greek the left sidebar's width is not enough.
Administration -> Stations -> Install GeoLite IP Database, translates into greek as 
"Εγκατάσταση Βάσης Δεδομένων GeoLite IP" 
which is longer than the sidebar's width.
And hence, part of it is hidden.
Here is a screenshot of the issue:

azuracast_greek

I tried investigating a bit more, but my knowledge in php, vue and ts and extremely limited (...zero!).
If this helps, two files seem to me related to this issue:

frontend/components/Admin/Stations.vue
frontend/components/Admin/menu.ts

Maybe, something like "auto width" could fix this.
Or, if the width needs to stay as is, different display of the translation???

### Relevant log output

_No response_

Vaalyn:
**Fixes issue:**
Fixes #7367

**Proposed changes:**
This PR aims to fix the problem with overlong link texts in the sidebar in other languages. For this I've added some classes on the nav items & links that add ellipsis when the text would overflow.

There is currently one problem that still needs to be fixed there, namely that I can't set the `title` attribute on `router-link` in order to get tooltips to work on those elements which is why this is a draft PR for now.

### DIFF
--- a/frontend/components/Common/SidebarMenu.vue
+++ b/frontend/components/Common/SidebarMenu.vue
@@ -51,21 +51,21 @@
                         <router-link
                             v-if="isRouteLink(item)"
                             :to="item.url"
-                            class="nav-link might-overflow ps-4 py-2"
+                            class="nav-link ps-4 py-2"
                             :class="getLinkClass(item)"
                             :title="item.title"
                         >
-                            {{ item.label }}
+                            <span class="might-overflow">{{ item.label }}</span>
                         </router-link>
                         <a
                             v-else
-                            class="nav-link might-overflow ps-4 py-2"
+                            class="nav-link ps-4 py-2"
                             :class="item.class"
                             :href="item.url"
                             :target="(item.external) ? '_blank' : ''"
                             :title="item.title"
                         >
-                            {{ item.label }}
+                            <span class="might-overflow">{{ item.label }}</span>
                             <icon
                                 v-if="item.external"
                                 class="sm ms-2"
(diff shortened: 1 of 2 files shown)


BusterNeece:
@Vaalyn I believe we have a class for this in our custom SCSS already, `.might-overflow` or something very similar...it's how we handle the same situation on the public player pages, so if you hover it unfurls the text but by default it uses ellipsis.

Vaalyn:
Have reused the existing `might-overflow` mixin as recommended and added this to the elements in the SideMenu, this is the result:

image
image
image

Vaalyn:
The additional `<span>`'s are because otherwise the icons vanish.

BusterNeece:
Hmm. That top icon doesn't look quite right. If you wanna keep hammering at it, that's fine, or just let me know and I can take a crack at it.

Vaalyn:
🤔 that's weird... the icon `<svg>` element has the class `.icon` which gives it a fixed width:
```css
width: 1em;
height: 1em;
display: inline-block;
...
```

But when the text is too long and thus overflows it shrinks... how? 😅 

I've added a `min-width` with the same `1em` as the `width` and that fixes it, not sure why that's needed though...

image

image

Vaalyn:
I think that's fine now:

image
image
image

BusterNeece:
Yeah that looks good to me 👍

PR #7388, opened 2024-08-13
Features: Maximum Bitrate, Maximum Mounts and Maximum HLS Streams

Yahav:
**This adds the ability for administrators to set:**

**A maximum bitrate for a station.**
This will limit the bitrates of:

- Mount points
- HLS Streams
- Remote Relays
- Live Streams Recordings

It includes a backend validation as well as a mechanism in which when downgrading a station bitrate limit it also sets any of the above that is higher than the limit to the new limit.
It also utilizes Icecast-KH limit-rate functionality to limit live streaming bitrate as well, in case one was to broadcast at a higher bitrate the stream will be "skippy"/"glichy".
Requested at: https://features.azuracast.com/suggestions/113370/station-bitrate-limits

**A maximum number of of mounts points and hls streams for a station.**
It includes a mechanism that will remove excessive mounts and hls streams when an admin change the limits to a lower one than the user currently have.

TODO (Could use some help here):

- [ ]  Should probably add a notice at the "Broadcasting Service" pane as well as the "Streamer/DJ Accounts" > "Connection Information" pane letting the user know his maximum bitrate for live streams.
- [ ]   As with any project there are standard and practices that i probably didn't met, this will require a thorough review.
- [ ]   Not sure how to deal with the translations, not sure how do i update the translation files with the new phrases.


Vaalyn:
I'm taking a look at the code in small bites for now so that this doesn't have to sit without any response until there is more time available to me. Hope that's okay since I'll not be able to gather every bit of feedback at once and will probably post a comment here and there in between 😅 

I like the naming of the `reduceMountPointsToLimit` & `reduceHlsStreamsToLimit` methods as they make the intent behind them explicitly clear just from the naming. Could we adjust the other methods to the same naming scheme?

- `lowerMountsBitrate` => `reduceMountsBitrateToLimit`
- `lowerHlsBitrate` => `reduceHlsBitrateToLimit`
- `lowerRemoteRelayAutoDjBitrate` => `reduceRemoteRelayAutoDjBitrateToLimit`
- `lowerLiveBroadcastRecordingBitrate` => `reduceLiveBroadcastRecordingBitrateToLimit`

backend/src/Controller/Api/Admin/StationsController.php
```diff
@@ -337,7 +339,38 @@ protected function handleEdit(Station $station): Station
...
+        }
+
+        $maxHlsStreamsLowered = $station->getMaxHlsStreams() !== 0
+            && ($oldMaxHlsStreams > $station->getMaxHlsStreams());
+        if ($maxHlsStreamsLowered) {
+            $this->stationRepo->reduceHlsStreamsToLimit($station);
+        }
+
+        if (
+            $adapterChanged || $maxBitrateChanged || $enabledChanged
```
Vaalyn, line 371: Would recommend to reformat this as follows for easier readability:
```php
if (
    $adapterChanged
    || $maxBitrateChanged
    || $enabledChanged
    || $maxMountsLowered
    || $maxHlsStreamsLowered
) {
```

backend/src/Controller/Api/Stations/HlsStreamsController.php
```diff
@@ -135,4 +137,16 @@ final class HlsStreamsController extends AbstractStationApiCrudController
 {
     protected string $entityClass = StationHlsStream::class;
     protected string $resourceRouteName = 'api:stations:hls_stream';
+
+    protected function createRecord(ServerRequest $request, array $data): object
+    {
+        $station = $request->getStation();
+        if ($station->getMaxHlsStreams() !== 0 && $station->getMaxHlsStreams() <= $station->getHlsStreams()->count()) {
+            throw new ValidationException(
+                'Unable to create a new stream, station\'s maximum HLS streams reached.'
```
Vaalyn, line 146: Haven't checked but those are shown in the UI as error popup messages, right?
Then we should make them translatable via: `__('Unable to create a new stream, station\'s maximum HLS streams reached.')`

Same applies for the other `ValidationException`'s

backend/src/Entity/Repository/StationRepository.php
```diff
@@ -131,7 +132,7 @@ public function resetHls(Station $station): void
             $streams = [
                 'aac_lofi' => 48,
                 'aac_midfi' => 96,
-                'aac_hifi' => 192,
+                'aac_hifi' => $station->getMaxBitrate() !== 0 ? $station->getMaxBitrate() : 128,
```
Vaalyn, line 135: The default should probably be kept at `192` instead of `128` if no max given

backend/src/Entity/Repository/StationRepository.php
```diff
@@ -147,6 +148,90 @@ public function resetHls(Station $station): void
...
+    {
+        foreach ($station->getMounts() as $mount) {
+            if ($mount->getAutodjBitrate() > $station->getMaxBitrate()) {
+                $mount->setAutodjBitrate($station->getMaxBitrate());
+                $this->em->persist($mount);
+            }
+        }
+
+        $this->em->flush();
+        $this->em->refresh($station);
```
Vaalyn, line 161: I don't think a refresh is necessary here since we only manipulate entities that are already tracked by Doctrine as part of the `Station` entity so it should know about those changes already. If I'm not misremembering a refresh would only be needed for example if a new entity was added with a relation to the station.

Vaalyn: The same should apply to the other new methods here

backend/src/Entity/Station.php
```diff
@@ -150,7 +154,7 @@ class Station implements Stringable, IdentifiableEntityInterface
     protected ?string $genre = null;
 
     #[
-        OA\Property(example: "/var/azuracast/stations/azuratest_radio"),
+        OA\Property(example: "/var/azuracast/stations/azuracast_radio"),
```
Vaalyn, line 157: Probably an unintended change

backend/src/Entity/Station.php
```diff
@@ -885,6 +919,53 @@ public function setTimezone(?string $timezone): void
         $this->timezone = $timezone;
     }
 
+    public function getMaxBitrate(): int
+    {
+        return $this->max_bitrate;
+    }
+
+    public function setMaxBitrate(int $max_bitrate): void
```
Vaalyn, line 927: Please use camelCase instead of snake_case for the parameters in the new setters here to adhere to PSR-12.

backend/src/Radio/Frontend/Icecast.php
```diff
@@ -189,6 +189,11 @@ public function getCurrentConfiguration(Station $station): ?string
                 'listenurl' => $this->getUrlForMount($station, $mountRow),
             ];
 
+            if ($station->getMaxBitrate() !== 0) {
+                $maxBitrateInBps = (int) $station->getMaxBitrate() * 1000 + 2500;
```
Vaalyn, line 193: Clarification questions:

- Is base `1000` instead of `1024` correct here?
- What's the extra `2500` for, is that to take small variance in the bitrate into account?

backend/src/Validator/Constraints/StationMaxBitrateChecker.php
```diff
@@ -0,0 +1,36 @@
...
+
+#[Attribute(Attribute::TARGET_PROPERTY)]
+final class StationMaxBitrateChecker extends Constraint
+{
+    public function __construct(
+        public string $stationGetter = '',
+        public string|array $selectedBitrate = '',
+        array $options = []
+    ) {
+
```
Vaalyn, line 18: Unnecessary empty line at beginning of method

backend/src/Controller/Api/Admin/StationsController.php
Yahav: Done

backend/src/Controller/Api/Stations/HlsStreamsController.php
Yahav: Done

backend/src/Entity/Repository/StationRepository.php
Yahav: I'm rather not comfortable enough to make this change :) 
could you help with this?

backend/src/Entity/Station.php
Yahav: Done

backend/src/Entity/Station.php
Yahav: Done

backend/src/Radio/Frontend/Icecast.php
Yahav: >Is base 1000 instead of 1024 correct here?
Fixed

>What's the extra 2500 for, is that to take small variance in the bitrate into account?
Exactly

backend/src/Validator/Constraints/StationMaxBitrateChecker.php
Yahav: Done

backend/src/Validator/Constraints/StationMaxBitrateChecker.php
```diff
@@ -0,0 +1,36 @@
...
+        $options['stationGetter'] = $stationGetter;
+        $options['selectedBitrate'] = $selectedBitrate;
+
+        parent::__construct($options);
+    }
+    public function getTargets(): string
```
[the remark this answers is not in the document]
Yahav: Done

Vaalyn:
> @Vaalyn all done expect for this Doctrine refresh one

I've just pushed a few commits (separated them by intent so they can be easier followed) 👍 

Could you verify if everything still works as expected for you too?

Yahav:
>Could you verify if everything still works as expected for you too?

Amazing. seem to work :)

PR #7391, opened 2024-08-15
OpenAPI.yml errors resolved

Lexian-droid:
**Fixes issue:**

As far as I know, no one has made an issue relating to this specific problem.

**Proposed changes:**

- Fixes OpenAPI structural errors and semantic errors.

These errors include:
```
Semantic error at paths./station/{station_id}/art/{media_id}
Declared path parameter "media_id" needs to be defined within every operation in the path (missing in "post", "delete"), or moved to the path-level parameters object
line 788

Semantic error at paths./station/{station_id}/art/{media_id}.post.parameters.1.name
Path parameter "id" must have the corresponding {id} segment in the "/station/{station_id}/art/{media_id}" path
line 814

Semantic error at paths./station/{station_id}/art/{media_id}.delete.parameters.1.name
Path parameter "id" must have the corresponding {id} segment in the "/station/{station_id}/art/{media_id}" path
line 841

Structural error at paths./station/{station_id}/frontend/{action}.post.parameters.1.required
should be equal to one of the allowed values
allowedValues: true
line 2530

Structural error at paths./station/{station_id}/backend/{action}.post.parameters.1.required
should be equal to one of the allowed values
allowedValues: true
line 2556
```

**Why?**
I'm working on an API client for Azuracast, and I was using OpenAPI Generator, which had given errors because of the syntax used in the openapi.yml file, after updating the YML, it has fixed the generator, which I believe Azuracast should have the updated YML.

### DIFF
--- a/backend/src/Controller/Api/Stations/Art/DeleteArtAction.php
+++ b/backend/src/Controller/Api/Stations/Art/DeleteArtAction.php
@@ -22,7 +22,7 @@ use Psr\Http\Message\ResponseInterface;
     parameters: [
         new OA\Parameter(ref: OpenApi::REF_STATION_ID_REQUIRED),
         new OA\Parameter(
-            name: 'id',
+            name: 'media_id',
             description: 'Media ID',
             in: 'path',
             required: true,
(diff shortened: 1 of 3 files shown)


Vaalyn:
As mentioned on Discord, we can't merge the PR atm. since the changes will need to be done on the respective `OA` Attributes in the code which is used to generate the openapi.yml.

One example, for the parameter rename from `id` to `media_id` on the delete endpoint with the `operationId: deleteMediaArt` that would need to be done here: https://github.com/AzuraCast/AzuraCast/blob/main/backend/src/Controller/Api/Stations/Art/DeleteArtAction.php#L25

PR #7400, opened 2024-08-19
Update to LS RR d32d509 - Fix LS config check

RM-FM:
This LS RR version fixes an issue with `liquidsoap --check` and stdin used by Azuracast config builder


RM-FM:
@BusterNeece fixes below exception when trying so save LS config in LS 2.3.x

Bildschirmfoto 2024-08-19 um 19 40 40

PR #7401, opened 2024-08-20
Features: playlist priorities and request control

caturria:
**Fixes issue:**

N/A
**Proposed changes:**

Adds prioritization to scheduled playlists, as well as features for controlling when requests can be played.

# Playlists

* Adds 'priority' field to station_playlists table;
* When multiple playlists are available, only those with the highest priority are considered.
* When multiple playlists are available at same priority, they are shuffled based on weights as usual.

## Backward Compatibility

* The priority field is optional
* If none of a station's playlists have priority configured (as is the case with existing installations), the exact behaviour of prioritizing based on playlist type is maintained;
* Priority should be considered all or nothing (playlists with unset priority will be interpreted as 0, which could make them unreachable).

# Request Control

I implemented controls which can be used to restrict the types of requests which can be played and when. Not only does this help to maintain the format of a show (by delaying irrelevant requests), it is also an abuse mitigation: it prevents the request system from being used to obstruct playback of scheduled programming such as ads.

## Request Priority

* A 'request' priority field has been added to the station table;
* If unset, the original behaviour of allowing requests at all times is preserved;
* If set, requests remain in queue until such time as the currently available playlists are at or below request priority.

## Requests Follow Station Format

* Added a 'requests_follow_format' field to station table;
* If set, requests can only play when a playlist they belong to is chosen;
* This allows requests to be played in a manner that does not disrupt the feel of the station;
* When a playlist is chosen for playback, if a request for media on that playlist is queued, it will take precedence over the normal rotation of that playlist;
* Requests in this mode even respect weight rules.
 
# Notes

I believe this represents a significant improvement in the capabilities of the auto DJ. This is largely a rewrite of the QueueBuilder class, and though it looks quite a bit different, it provides flexibility beyond what was previously achievable. As this implementation works request processing directly into the auto DJ calculations, it removes the need for the ProcessRequests event.

# Frontend Concerns

As previously disclosed, I am blind, and cannot competently implement these controls on the frontend. I managed to get priority to display on the playlists form by using weight as a guide, but I cannot confirm whether the appearance is correct. I could not get it to display in the playlists table, however.
If you would be willing to assist with this portion, here's what's needed:

* Playlists page: display priority;
* Playlists form: allow configuration of priority (my attempt at this is in PR);
* Station configuration (song requests tab) needs Request Priority and "Requests must follow station format" options.

I am of course available to address any issues that might arise with this. Finally, if there are style issues that I missed or was unaware of, please feel free to point those out as well.


caturria:
Will commit to resolve test failures in the morning.

BusterNeece:
@caturria Handy tip for your dev work: if you want to replicate the code formatting and test suite locally, you can just run `make bash` to shell into the container, then `composer run cleanup-and-test`, which will both auto-fix any styling issues it finds and will identify the same errors that our CI finds here.

caturria:
@BusterNeece that is very helpful, thank you.

I did have a couple more questions for you about development workflow.

Is there a command (or at least some easier way) to extrapolate only PHP errors from the logs? This has been the most frustrating part of debugging this: finding a relevant tidbit amongst all the noise has been difficult. Sometimes I can grep out certain phrases (.ERROR, .WARNING, QueueBuilder.php: etc) but sometimes I'm not so lucky and have to listen through an unbelievable amount of irrelivant data to find a single line that's relevant to me.

If PHP errors persist for long enough to cause all the workers to die, is there an easy way to start them all back up that's less disruptive than restarting the containers?

BusterNeece:
@caturria Just the PHP error messages are logged to `/var/azuracast/www_tmp/app.log` (though it rotates, so it'll actually be `app-yyyy-mm-dd.log`, but that should help with filtering them out specifically.

There's also a service control on the Administration dashboard where you can specifically restart `php-worker` without needing to restart the whole container.

caturria:
Thanks @BusterNeece!

I believe the PR is ready for review; test pass, the inevitable dumb regression I introduced while addressing the tests is fixed and it seems to be working great.

One other question for you. Is there an established way to install a production instance from a personal fork/ branch?

BusterNeece:
@caturria Only if it's a dev branch that we maintain, i.e. our Liquidsoap 2.3.x test branch. Otherwise, you'd have to enable GitHub Actions on the repo and then push your own docker images up, then update `docker-compose.override.yml` to reflect those images instead. It'd be a bit of work.

BusterNeece:
@caturria Happy to review this whenever, just let me know when you think it's in a good state.

caturria:
@BusterNeece I think I can safely say go for it.

During testing I found more edge cases that caused incomplete playback of sequences (especially shortly after a station restart). I think I finally have those weeded out. My instance has been correctly playing out sequenced events for the past six hours.

I managed to get at least some frontend done, but I of course can't vouch for what it looks like. I didn't commit it, but I got 'requests follow station format' to appear on the requests form. The checkbox populates with the correct value from the API but doesn't save. I was entirely unsuccessful in getting 'request priority' to show up though.
Vue is not a technology I'm versed in.

BusterNeece:
@caturria Just so you know I am actively working on reviewing this branch locally, so please try to avoid committing anything tonight if possible.

caturria:
Hi @BusterNeece
No problem; the only thing I'm committing to tonight is sleep.
The only thing I did was merged in latest main.

caturria:
Hi @BusterNeece,

Would you support the idea of having another class created to hold that state (QueueBuilderState or some such) that is instantiated locally (in CalculateNextSongs) and passed about?

That would satisfy your concerns about possible concurrent use of the same instance while still keeping things contained and (IMHO) easy to reason about.

caturria:
Hi @BusterNeece,

Please disregard my earlier message.
Please see the latest commit. I refactored it to eliminate the previous internal state as requested. I also removed the conditional eligibility checks I had added as they did not save a consequential amount of time on checks and made the logs confusing.

caturria:
Hi @BusterNeece,

Thank you very much for your support in getting this in.

Could I please trouble you for some assistance with the frontend though?
Request priority and 'Requests Must Follow Station Format' need to be added to the 'Song Requests' tab of station profile.

Thank you.

BusterNeece:
@caturria Looking into this...can you clarify for me what you intended with the "Requests must follow station format" setting?

It seems it's only used in one single place to prevent general requests from being submitted in the AutoDJ queue builder, but I don't see anywhere that it's used to _allow_ requests through.

caturria:
Hi @BusterNeece,

It forces requests to wait until a playlist containing the requested song is chosen by the auto DJ.
The code that ultimately lets requests go through in this mode begins around line 369 of QueueBuilder.php:
```
            //Try to play requests for this playlist first, then try to play it normally.
            if (
                null === $request
                && !$playlist->backendMerge()
            ) {
                $request = $this->getRequestForPlaylist(
                    $playlist,
                    $requests,
                    $expectedPlayTime
                );
            }

            if (null !== $request) {
                return $this->playRequest(
                    $request,
                    $expectedPlayTime
                );
            }
```

This logic is not specific to this setting, it simply puts requests for the chosen playlist's contents through before performing standard track selection from it. When requests follow station format is on, this is the only way for a request to process.

BusterNeece:
@caturria Got it! I added the frontend config settings for the station, and also did some tweaking to the logic of the QueueBuilder class so it uses a generator instead of direct array walking manipulation; should help a little with code readability and logic.

PR #7425, opened 2024-09-03
More scheduling anomaly fixes

caturria:
**Fixes issue:**

**Proposed changes:**

Hopefully the last batch of significant adjustments to scheduling internals needed to resolve edge cases.

# Prevents eligibility checks from going back and forth in time

If a scheduled playlist began with a very short track (such as a 'We'll be right back' message, a random unscheduled song could sometimes be inserted in-between this and subsequent items in the scheduled playlist. This is similar in nature to the previous validation issue, in that this happens when the playout runs ahead of schedule such that the drift between expected and actual playtime is enough to make a track ineligible after already being scheduled. The system would check playlist eligibility for a time earlier than a previous schedule check, causing a playlist to rapidly go back and forth between being active and inactive.

## New fields added to station queue

* schedule_id
* timestamp_scheduled
* is_cancelled (see below).

## Necessary adjustments to scheduler internals

I added new out parameters to internal scheduler functions so that extra information (namely the specific schedule item that triggered a playlist as well as the qualifying date range) could propagate through the scheduler and be exported.

Now, when deciding what time to check eligibility for, the system will use either the current 'Expected play time' or the time of the last scheduled track, whichever is later. The timestamp_scheduled field holds either the start timestamp of the associated schedule items, or the original expectedPlayTime (for unscheduled items). This also makes 'has played since' checks more accurate.

# Soft-deletable queue items

I added the ability for queue items to be 'cancelled', or soft-deleted. If you perform a manual delete action on a queue item (or 'Clear upcoming song queue'), it will mark the items as cancelled instead of deleting. This was done for the following reasons:
* Prevents scheduled items from being immediately requeued if manually deleted by admin
* Will make investigation of future edge cases easier with some evidence left behind
* Paves the way for some of the advanced features I would like to contribute in the near future.

This finally solves all of the scheduling problems I have personally experienced. As usual, I'm sure there are some issues with spacing and the like which the test suite didn't catch. Thank you once again for this wonderful software.

### DIFF
--- a/backend/src/Controller/Api/Stations/Files/BatchAction.php
+++ b/backend/src/Controller/Api/Stations/Files/BatchAction.php
@@ -362,6 +362,7 @@ final class BatchAction implements SingleActionInterface
 
                     $newQueue = StationQueue::fromMedia($station, $media);
                     $newQueue->setTimestampCued($cuedTimestamp);
+                    $newQueue->setTimestampScheduled($cuedTimestamp);
                     $newQueue->setIsPlayed();
                     $this->em->persist($newQueue);
 
--- a/backend/src/Controller/Api/Stations/Files/BatchAction.php
+++ b/backend/src/Controller/Api/Stations/Files/BatchAction.php
@@ -310,6 +310,7 @@ final class BatchAction implements SingleActionInterface
 
                     $newQueue = StationQueue::fromMedia($stationRef, $media);
                     $newQueue->setTimestampCued($cuedTimestamp);
+                    $newQueue->setTimestampScheduled($cuedTimestamp);
                     $this->em->persist($newQueue);
                 } catch (Throwable $e) {
                     $result->errors[] = sprintf('%s: %s', $media->getPath(), $e->getMessage());
--- a/backend/src/Controller/Api/Stations/QueueController.php
+++ b/backend/src/Controller/Api/Stations/QueueController.php
@@ -14,6 +14,7 @@ use App\Http\ServerRequest;
 use App\OpenApi;
 use App\Radio\AutoDJ\Queue;
 use App\Utilities\Types;
+use InvalidArgumentException;
 use OpenApi\Attributes as OA;
 use Psr\Http\Message\ResponseInterface;
 use Symfony\Component\Serializer\Serializer;
@@ -173,5 +174,19 @@ final class QueueController extends AbstractStationApiCrudController
         $this->queueRepo->clearUpcomingQueue($station);
 
         return $response->withJson(Status::deleted());
+    }
+
+            /**
+         * Don't delete, just cancel.
+         * This prevents playout, but keeps the record around for other decision making, bug investigation, etc.
+         */
+    protected function deleteRecord(object $record): void
+    {
+        if (!($record instanceof $this->entityClass)) {
+            throw new InvalidArgumentException(sprintf('Record must be an instance of %s.', $this->entityClass));
+        }
+        $record->setIsCancelled(true);
+        $this->em->persist($record);
+        $this->em->flush();
     }
 }
--- a/backend/src/Controller/Api/Stations/QueueController.php
+++ b/backend/src/Controller/Api/Stations/QueueController.php
@@ -176,16 +176,18 @@ final class QueueController extends AbstractStationApiCrudController
         return $response->withJson(Status::deleted());
     }
 
-            /**
-         * Don't delete, just cancel.
-         * This prevents playout, but keeps the record around for other decision making, bug investigation, etc.
-         */
+    /**
+     * Don't delete, just cancel.
+     * This prevents playout, but keeps the record around for other decision making, bug investigation, etc.
+     */
     protected function deleteRecord(object $record): void
     {
         if (!($record instanceof $this->entityClass)) {
             throw new InvalidArgumentException(sprintf('Record must be an instance of %s.', $this->entityClass));
         }
+
         $record->setIsCancelled(true);
+
         $this->em->persist($record);
         $this->em->flush();
     }
--- a/backend/src/Controller/Api/Stations/Requests/ListAction.php
+++ b/backend/src/Controller/Api/Stations/Requests/ListAction.php
@@ -119,10 +119,11 @@ final class ListAction extends AbstractSearchableListAction
             foreach ($playlists as $playlist) {
                 if (
                     $this->scheduler->isPlaylistScheduledToPlayNow(
-                        $this->scheduler->createContext()
-                        ->withPlaylist($playlist)
-                        ->withNow($now)
-                        ->withExcludeSpecialRules(true)
+                        new SchedulerContext(
+                            $playlist,
+                            $now,
+                            true
+                        )
                     )
                 ) {
                     $ids[] = $playlist->getIdRequired();
--- a/backend/src/Controller/Api/Stations/Requests/ListAction.php
+++ b/backend/src/Controller/Api/Stations/Requests/ListAction.php
@@ -15,6 +15,7 @@ use App\Http\Response;
 use App\Http\ServerRequest;
 use App\OpenApi;
 use App\Radio\AutoDJ\Scheduler;
+use App\Radio\AutoDJ\SchedulerContext;
 use Carbon\CarbonImmutable;
 use OpenApi\Attributes as OA;
 use Psr\Cache\CacheItemPoolInterface;
@@ -116,7 +117,14 @@ final class ListAction extends AbstractSearchableListAction
 
             /** @var StationPlaylist $playlist */
             foreach ($playlists as $playlist) {
-                if ($this->scheduler->isPlaylistScheduledToPlayNow($playlist, $now, true)) {
+                if (
+                    $this->scheduler->isPlaylistScheduledToPlayNow(
+                        $this->scheduler->createContext()
+                        ->withPlaylist($playlist)
+                        ->withNow($now)
+                        ->withExcludeSpecialRules(true)
+                    )
+                ) {
                     $ids[] = $playlist->getIdRequired();
                 }
             }
--- a/backend/src/Entity/Migration/Version20240901011513.php
+++ b/backend/src/Entity/Migration/Version20240901011513.php
@@ -17,15 +17,17 @@ final class Version20240901011513 extends AbstractMigration
     {
         $this->addSql('ALTER TABLE station_queue ADD timestamp_scheduled INT, ADD schedule_id INT DEFAULT NULL');
         $this->addSql('CREATE INDEX idx_timestamp_scheduled ON station_queue (timestamp_scheduled)');
+        $this->addSql('ALTER TABLE station_queue ADD CONSTRAINT FK_277B0055A40BC2D5 FOREIGN KEY (schedule_id) REFERENCES station_schedules (id) ON DELETE CASCADE');
+        $this->addSql('CREATE INDEX IDX_277B0055A40BC2D5 ON station_queue (schedule_id)');
         $this->addSQL('UPDATE station_queue SET timestamp_scheduled = timestamp_played');
         $this->addSQL('ALTER TABLE station_queue CHANGE timestamp_scheduled timestamp_scheduled INT NOT NULL');
     }
 
     public function down(Schema $schema): void
     {
-        $this->addSql('ALTER TABLE station CHANGE max_bitrate max_bitrate INT DEFAULT 0, CHANGE max_mounts max_mounts TINYINT(1) DEFAULT 0, CHANGE max_hls_streams max_hls_streams TINYINT(1) DEFAULT 0');
+        $this->addSql('ALTER TABLE station_queue DROP FOREIGN KEY FK_277B0055A40BC2D5');
+        $this->addSql('DROP INDEX IDX_277B0055A40BC2D5 ON station_queue');
         $this->addSql('DROP INDEX idx_timestamp_scheduled ON station_queue');
         $this->addSql('ALTER TABLE station_queue DROP timestamp_scheduled, DROP schedule_id');
     }
-
 }
--- a/backend/src/Entity/Migration/Version20240901011513.php
+++ b/backend/src/Entity/Migration/Version20240901011513.php
@@ -10,7 +10,7 @@ final class Version20240901011513 extends AbstractMigration
 {
     public function getDescription(): string
     {
-        return '';
+        return 'Track timestamp_scheduled and schedule_id in station_queue.';
     }
 
     public function up(Schema $schema): void
--- a/backend/src/Entity/Migration/Version20240901011513.php
+++ b/backend/src/Entity/Migration/Version20240901011513.php
@@ -1,0 +1,31 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Entity\Migration;
+
+use Doctrine\DBAL\Schema\Schema;
+
+final class Version20240901011513 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return '';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE station_queue ADD timestamp_scheduled INT, ADD schedule_id INT DEFAULT NULL');
+        $this->addSql('CREATE INDEX idx_timestamp_scheduled ON station_queue (timestamp_scheduled)');
+        $this->addSQL('UPDATE station_queue SET timestamp_scheduled = timestamp_played');
+        $this->addSQL('ALTER TABLE station_queue CHANGE timestamp_scheduled timestamp_scheduled INT NOT NULL');
+    }
+
+    public function down(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE station CHANGE max_bitrate max_bitrate INT DEFAULT 0, CHANGE max_mounts max_mounts TINYINT(1) DEFAULT 0, CHANGE max_hls_streams max_hls_streams TINYINT(1) DEFAULT 0');
+        $this->addSql('DROP INDEX idx_timestamp_scheduled ON station_queue');
+        $this->addSql('ALTER TABLE station_queue DROP timestamp_scheduled, DROP schedule_id');
+    }
+
+}
--- a/backend/src/Entity/Migration/Version20240902155213.php
+++ b/backend/src/Entity/Migration/Version20240902155213.php
@@ -1,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Entity\Migration;
+
+use Doctrine\DBAL\Schema\Schema;
+
+final class Version20240902155213 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'Add is_cancelled field to station_queue';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE station_queue ADD is_cancelled TINYINT(1) NOT NULL DEFAULT 0');
+        $this->addSql('CREATE INDEX idx_is_cancelled ON station_queue (is_cancelled)');
+    }
+
+    public function down(Schema $schema): void
+    {
+        $this->addSql('DROP INDEX idx_is_cancelled ON station_queue');
+        $this->addSql('ALTER TABLE station_queue DROP is_cancelled');
+    }
+}
--- a/backend/src/Entity/Repository/StationQueueRepository.php
+++ b/backend/src/Entity/Repository/StationQueueRepository.php
@@ -8,6 +8,7 @@ use App\Entity\Interfaces\SongInterface;
 use App\Entity\Station;
 use App\Entity\StationMedia;
 use App\Entity\StationPlaylist;
+use App\Entity\StationSchedule;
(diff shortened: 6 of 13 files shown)


BusterNeece:
@caturria I'm a little torn on some of this work...I already mentally have a hard time grasping the entire ruleset of the AutoDJ scheduler in my head at once, and my ability to do that is essential to being able to diagnose and debug issues. This introduces a fair amount of extra complexity into the process, and honestly just reviewing it I found myself zoning out a few times.

I'm not saying that we necessarily need to be coding specifically to my ability (or lack thereof), but the simpler we can keep this essential code, the better.

One thing I will say I don't like from a "code smell" perspective is the passed-by-reference arrays thing that keeps popping up in these new PRs. It's not always clean to return an array of arrays, but the cognitive load of pass-by-reference arrays is pretty high, and it honestly makes it much harder for me to trace the lifecycle of a variable just by looking at it. I'd like to try avoiding it if at all possible in our code.

caturria:
Hi @BusterNeece,

Thank you for your feedback.

First and foremost, I wanted to thank you for the solid collaboration we've had so far.

I think that we're clashing a little bit over coding style, because we have different backgrounds and are a custom to different things. I'm sure it's nothing a little dialogue can't sort out, and I'm always open to suggestions that could make this easier on both of us.

I understand that there may be contrast between what causes high cognitive load for a blind programmer as compared to a sighted one. For me, the primary trigger is deeply nested control structures. Since I can't go by indentation, I have to count braces in my head when things go wrong, so I tend to start thinking about refactoring code into multiple functions when it goes beyond three levels deep.

To address your concerns about pass-by-reference, this is a tool I've been reaching for in order to keep the scheduling classes stateless, as previously requested. I would have preferred to simply include getScheduleItem() and getDateRange() methods so the caller could fetch this extra qualifying information after an eligibility check, rather than having to carry it all the way through the mechanism by reference. You said that you did not want state in these classes due to concerns about concurrency, but these objects are transient: they live to schedule one song and that's it. Even if multiple stations were processing their queues concurrently, they would not share a single instance of either the scheduler or the queue builder.

If I can't put state in the classes that the caller can pull out, and I can't pass it through by reference (which admittedly feels unnatural), then with all due respect, the toolbox is starting to feel a little empty. If you have a preference for a particular technique in these instances, I'm more than happy to adapt.

Finally, I would like to address your comments about simplicity. My use case for AzuraCast involves scheduling about a hundred short, automatically generated playlists per day, which pushed the scheduler beyond its limits. As we've discussed, I encountered a number of those hard-to-diagnose, intermittent anomalies that had me reaching for custom Liquidsoap and other workarounds. With this PR (as well as the last few that have already been merged), I've tackled all of them.

Finally, I'm sure I'm not alone in desiring some additional "power user" features that go beyond just 'play one of these per hour' or 'play this from time A to time B'. I have the passion, the time, and (I believe) the no-how to implement these if we can figure out how to best complement each others' strengths. As for debugging, I'm always just a mention away if I don't catch an issue myself when it pops up, and I'm happy to help out in any way I can.

Please advise.

BusterNeece:
@caturria I think for this particular purpose, we are going to need a DTO object. I'm taking a crack at it right now and will update accordingly.

BusterNeece:
@caturria I feel like the optimal solution here might be to split `Scheduler` into two components, one being the stateless utility functions that we can dependency-inject across the app, and the other being a specific `ScheduledPlaylist` class, which _is_ stateful and represents a single playlist as we're handling it and tossing it around.

`QueueBuilder` shouldn't carry with it the state of a single playlist that it's dealing with, as it iterates through several playlists in the process of determining which should play and this would get extremely messy. Likewise, `Scheduler` has utility classes that are certainly called by multiple places in the application.

If that works for you, I can look into implementing this, or if you'd like to take a crack at it, you're welcome to.

caturria:
Would you be comfortable with a single-use context object that the principal functions can take in and operate on? Something like this:
```
$ctx = new SchedulingContext(
$expectedPlayTime,
...);
if ($scheduler->shouldPlaylistPlayNow($ctx)) {
//...
$dateRange = $ctx->getDateRange();
}
```
Keeps state in local scope, keeps it contained, and keeps function signatures short.

I could create a similar context object for the QueueBuilder, which would nicely avoid having to pass all that junk by reference all over the place.
Thoughts?

BusterNeece:
@caturria Yep, that's perfectly fine.

As a matter of principle, I just avoid mixing Dependency-Injected classes (`Scheduler` and `QueueBuilder` both being that) with state preservation (i.e. entities), but context classes that _are_ stateful and get tossed around instead of by-reference arrays are totally fine by me.

caturria:
Cool. Thank you.

I didn't like the pass-by-reference arrays either, but I guess I mistook you as being opposed to using stateful objects in general.

I'll be back with an update.

BusterNeece:
@caturria Sounds good. Sorry for the confusion...we definitely do use stateful objects in various places! It's just that by design, our dependency-injected classes are "singletons", which makes state tricky.

caturria:
Okay. I stand corrected. I assumed they were instantiated per request. My bad.

caturria:
Hi @BusterNeece,

I was hoping I could bounce a couple more stylistic questions off you.

First and foremost, I've noticed that you've often had to commit numerous spacing adjustments to my PRs, even after I've resolved all of the warnings issued by phpcs. I would like to try to save you as much of this work as possible, but spacing doesn't provide the same benefits to a blind dev as it does to a sighted one, and even noticing a spacing mistake is very difficult for me. Would you mind providing a few examples of the specific spacing issues you've had to manually address and why? I can't promise that I'll catch all of them, but I'd certainly feel better if I knew I at least caught more than I've been able to so far.

I still miss things even after reading over PSR-12 several times. For instance, I was under the impression that I should break up a function signature across multiple lines if it contains more than one argument. But then I come across a call ($a, $b, $c); one-liner somewhere, and realize that there must be nuance at play that will simply never be obvious to someone who doesn't experience the world visually.

Lastly, do you have an opinion on early returns VS. nested ifs?
When working on a function like this:
```
function foo(): bool
{
if (good) {
//remember: 1
//do stuff.
if (otherGood) {
//remember: 2
//do stuff.
}
//remember: 1
//Do other stuff.
foreach(...) {
//remember: 2
if (anotherGood) {
//remember: 3
// do stuff.
return true;
}
//remember: 2
}
//remember: 1
}
return false;
}
```
It is far easier for me if I can change it to:
```
function foo(): bool
{
if (bad) {
return false;
}
//do stuff.
if (otherBad) {
return false;
}
//Do other stuff.
foreach (...) {
//1
if (anotherBad) {
//2
continue;
}
//1
//do stuff.
}
return true;
}
```

Am I ruffling any feathers by doing this? Even eliminating just one level of nesting makes a huge difference to cognitive load when my only means of reconciling nested code is to keep a count of braces in my head. I'm curious to know how different types of control structures effect cognitive load for you (or any other sighted developer) and how they contrast with my experiences. I noticed that you tend to prefer array_walk() and friends with a lambda instead of a loop, which feels less natural to me.
Thank you very much.

BusterNeece:
@caturria To be honest, most of what I'm fixing isn't expressly specified in PSR-12, it's just sort of keeping the code cleaner for visual review. Most of my commits involve adding a space after comments (i.e. `// here`) and between the end of one function and the start of the next, for example.

In terms of whether functions get wrapped onto new lines, I usually base that on whether the combined one-liner would threaten to exceed the 120-character line width limit. If it's just two or three args and would easily fit on one line, I put it on one line.

As for the "return early" mentality, I agree with you! I haven't rewritten a majority of those items, because early returns and avoiding nesting are something I consider a good practice too.

caturria:
Hi @BusterNeece,

I've had to slow down on this quite a bit for the last couple of days due to other commitments. But I was looking at it a bit more tonight and honestly, I can't help but think that dependency injecting the scheduler (and possibly queue builder) is simply not the best approach. Treat these as transient objects that get instantiated by dependent classes as needed (or injected via factories) and we get simpler internals.

No more twisting ourselves into pretzels trying to brainstorm the least "ugly" way to do something that is inherently "ugly" (ferrying a growing number of individual state items in and out through several levels of plumbing). Existing parts of the code that are happy with the existing boolean answers about scheduling don't have to be changed, and those parts that need extra detail can just pass it in and out of the scheduler through simple gets and sets.

Is this something you could support?

BusterNeece:
@caturria Trust me, this is something I've wrestled with back and forth for quite some time on our codebase.

Basically, if we're dependency-injecting services into a class, then for us to be able to 'autowire' that class into other classes, the whole chain, top to bottom, has to be injectable (which, with the limitations of our DI library, means they have to be singletons).

If we make `Scheduler` have injected dependencies but not, itself, be autowirable (because it'd need to be instantiated via a factory `make` call, which does exist), then that would be one piece of code in a huge application that doesn't follow the rules of the entire rest of the application.

As I've said in previous PR comments, I'm not going to change the entire paradigm by which the code is built just for one portion of its functionality, like the scheduler. That reeks to me of tech debt.

I know it may not be as convenient to have a context object in this case, but I have to hold firm on that just for the sake of ensuring consistency across the codebase.

BusterNeece:
@caturria I will say, though, as an alternative...if you wanted to create a `SchedulerFactory` that _was_ dependency-injected and a singleton, and that just had a method like `createForPlaylist` that created a stateful instance and injected in the needed dependencies...that'd be fine. We have to do that one or two other places in the code and that works fine.

caturria:
Hi @BusterNeece,

Got some time today, so I'm taking a swing at this.

I haven't done any integration yet, but I committed a draft SchedulerContext object. Could you take a look and tell me if this is on track with what you have in mind? It provides a builder-style interface for the caller and callee to set inputs and outputs, respectively:
```
->withPlaylist($playlist)
->withExcludeSpecialRules(true)
);
$this->scheduler->shouldPlaylistPlayNow($ctx);
```

Also, one more quick question. I noticed this change you made around line 42 of QueueBuilder:
```
    private const array LEGACY_PRIORITIES = [
        PlaylistTypes::OncePerHour->value => 6,
        PlaylistTypes::OncePerXSongs->value => 4,
        PlaylistTypes::OncePerXMinutes->value => 2,
        PlaylistTypes::Standard->value => 0,
    ];
```
Specifically, the values [0, 1, 2, 3] were changed to [0, 2, 4, 6]. I'm just wondering what for, as this changes the rules, and introduces the possibility of unexpected ties if we were to ever add a new playlist type.
```
        $scheduled = count($playlist->getScheduleItems()) > 0 ? count(PlaylistTypes::cases()) : 0;
        return (self::LEGACY_PRIORITIES[$playlist->getType()->value] ?? 0) + $scheduled;
```

Here, a scheduled standard (5) is beaten by an unscheduled once per hour (6), but not by any other unscheduled once per (4 and 2).

Is it just that you want the numbers even? If so, the result of PlaylistTypes::cases() needs to be multiplied by two in order to maintain behaviour.

BusterNeece:
@caturria Will look at the new schedule context class when I can. Sounds fine from what you're describing, though.

Making the legacy priorities even was specifically to make sure that the rules matched the actual "legacy" rules. In the old scheduler, the priority is exactly what it winds up being now (type x scheduled, then type x unscheduled, then type y scheduled, then type y unscheduled, descending). With the numbers directly next to each other, a scheduled item of one type would match the priority of an unscheduled item of another type, which doesn't match legacy behavior.

BusterNeece:
@caturria Looking at the Schedule Context...you could just make those public properties. The only reason we don't just make Doctrine Entities public properties is because we often need to have "side-effects" with those method calls. It doesn't look like that'll be the case with this context object.

caturria:
Hi @BusterNeece,

I think we're ready for another iteration, and I do need clarification on a couple of things.

I now have the SchedulerContext of course, as well as a QueueBuilderContext.

I just got your note about public properties: I can make that switch if you prefer. But I do like the builder syntax, and the context ended up with a few methods like "getPlaylistRequired" which throw if the property they fetch isn't set. Perhaps we can cut down on preamble in some methods by destructuring public properties out of the context, but the extra null checks that would be required might just make it six of one, half dozen of the other. Thoughts?

The QueueBuilderContext now tracks the state of eligible playlists. You create the context, then pass each playlist to a registration method. The context itself does all the checks, then admits the playlist if it qualifies. Then just iterate: the context takes care of ordering.

Lastly, I need a little help with the test cases. I updated StationPlaylistTest to reflect that scheduler methods now take the context instead of several individual parameters. But I can't seem to get the tests to run. When I run 'composer run cleanup-and-test-full' or 'composer run codeception' from within the container, I get:
```
azuracast@a0e9653a24ba:~/www$ composer run codeception                                                                  

> codecept run --no-interaction --coverage --coverage-xml --fail-fast                                                   

Codeception PHP Testing Framework v5.1.2 https://stand-with-ukraine.pp.ua                                               

                                                                                                                        

In XdebugDriver.php line 145:                                                                                           

                                                                                                                        

  XDEBUG_MODE=coverage (environment variable) or xdebug.mode=coverage (PHP configuration setting) has to be set         

                                                                                                                        

                                                                                                                        

run [-o|--override OVERRIDE] [-e|--ext EXT] [--report] [--html [HTML]] [--xml [XML]] [--phpunit-xml [PHPUNIT-XML]] [--colors] [--no-colors] [--silent] [--steps]
[-d|--debug] [--shard SHARD] [--filter FILTER] [--grep GREP] [--bootstrap [BOOTSTRAP]] [--no-redirect] [--coverage [COVERAGE]] [--coverage-html [COVERAGE-HTML]]
[--coverage-xml [COVERAGE-XML]] [--coverage-text [COVERAGE-TEXT]] [--coverage-crap4j [COVERAGE-CRAP4J]] [--coverage-cobertura [COVERAGE-COBERTURA]] [--coverage-phpunit
[COVERAGE-PHPUNIT]] [--no-exit] [-g|--group GROUP] [-s|--skip SKIP] [-x|--skip-group SKIP-GROUP] [--env ENV] [-f|--fail-fast [FAIL-FAST]] [--no-rebuild]
[--seed SEED] [--no-artifacts] [--] [<suite> [<test>]]                       

                                                                                                                        

Script codecept run --no-interaction --coverage --coverage-xml --fail-fast handling the codeception event returned with error code 1                   
                                                                                        

azuracast@a0e9653a24ba:~/www$ export XDEBUG_MODE=coverage                                                               

azuracast@a0e9653a24ba:~/www$ composer run codeception                                                                  

> codecept run --no-interaction --coverage --coverage-xml --fail-fast                                                   

Codeception PHP Testing Framework v5.1.2 https://stand-with-ukraine.pp.ua                                               

                                                                                                                        

In Finder.php line 646:                                                                                                 

                                                                                                                        

  The "/var/azuracast/www/src" directory does not exist.                                                                

                                                                                                                        

                                                                                                                        

run [-o|--override OVERRIDE] [-e|--ext EXT] [--report] [--html [HTML]] [--xml [XML]] [--phpunit-xml [PHPUNIT-XML]] [--colors] [--no-colors] [--silent] [--steps] 
```

What is the appropriate command to run the test coverage?

StationPlaylistTest will currently fail due to an earlier change which made 'once per x' options use recently played songs as a source of truth instead of relying on played_at. I can take care of updating that if I can get the tests to run.

BusterNeece:
@caturria Yeah, the Codeception test suite hasn't worked in several months now. Some infrastructural changes we had to make to our CI completely broke it, and I haven't been actively keeping it up-to-date with our code changes, so it's likely things aren't working right on it at all right now. It's something on the bigger to-do list pile to update those test suites to be more frontend-focused (as the Vue 3 SPA rewrite really did a number on our ability to test things).

As for the properties on that context object...I'd honestly prefer to have those be public properties with the occasional method-call/private-property where appropriate (i.e. most of the properties could be public in this case, but ones that need more processing could follow the more Doctrine-esque model) just to avoid writing a ton of boilerplate. Also, we don't use the "builder" pattern basically anywhere else in the application right now so it would be a one-off in and of itself.

caturria:
Hi @BusterNeece,
That's been done.

I kept `getPlaylistRequired()` and `getExpectedPlayTimeRequired()`, because these are required to be set in all cases except for the streamer-related schedule checks.

Just to clarify, for the sake of brevity, is it cool to hold public properties in local vars when they're referenced several times in a single function, or should I be accessing them directly all the time?

BusterNeece:
@caturria Yeah that's fine.

caturria:
@BusterNeece thank you. This should be ready to go then.

caturria:
Hi @BusterNeece,

Thank you.

I'm so sorry to have to hit you up with something else so quickly.
I was just about to open a bug as I discovered yet another skipped track issue this aft. But this time it's not because of the scheduler.

If Liquidsoap tries to make an API call right at the beginning of (or sometimes right at the end of) a minute, there's a good chance it will exceed that two second timeout and come back with (522), because it's competing with all the once-per-minute sync tasks for a database lock.

Would you be okay with increasing the timeout at least as an interim solution?
I know it's supposed to time out and try again, but if this happens on a nextsong call, it's of course going to dequeue a track and lose it. If it happens to the feedback call, then it's going to miss updating now playing and song history.
Here are some logs: notice that most of these happen right at the beginning or end of a minute, except for one that went on for 30 seconds or so (I guess one of the longer periodic tasks was running).
```
2024/09/09 06:21:02 [lang:3] API feedback - Response (522):                                                                                            
                           

2024/09/09 14:31:03 [lang:3] API feedback - Response (522):                                                                                            
                           

2024/09/09 14:33:03 [lang:3] API nextsong - Response (522):                                                                                             
2024/09/08 04:50:04 [lang:3] API nextsong - Response (522):                                                                                            
                           

2024/09/08 04:50:06 [lang:3] API nextsong - Response (522):                                                                                            
                           

2024/09/08 04:50:07 [lang:3] API feedback - Response (522):                                                                                            
                           

2024/09/08 06:31:03 [lang:3] API feedback - Response (522):                                                                                            
                           

2024/09/08 10:55:02 [lang:3] API feedback - Response (522):                                                                                             2024/09/07 18:07:05 [lang:3] API feedback - Response (522):                                                                                            
                           

2024/09/07 21:45:02 [lang:3] API feedback - Response (522):                                                                                            
                           

2024/09/08 00:17:02 [lang:3] API nextsong - Response (522):                                                                                            
                           

2024/09/08 00:17:55 [lang:3] API nextsong - Response (522):                                                                                            
                           

2024/09/08 00:17:58 [lang:3] API nextsong - Response (522):                                                                                            
                           

2024/09/08 00:18:01 [lang:3] API nextsong - Response (522):                                                                                            
                           

2024/09/08 00:18:03 [lang:3] API nextsong - Response (522):                                                                                            
                           

2024/09/08 00:18:05 [lang:3] API nextsong - Response (522):                                                                                            
                           

2024/09/08 00:18:07 [lang:3] API nextsong - Response (522):                                                                                            
                           

2024/09/08 00:18:09 [lang:3] API nextsong - Response (522):                                                                                            
                           

2024/09/08 00:18:11 [lang:3] API nextsong - Response (522):                                                                                            
                           

2024/09/08 00:18:13 [lang:3] API nextsong - Response (522):                                                                                            
                           

2024/09/08 00:18:15 [lang:3] API nextsong - Response (522):                                                                                            
                           

2024/09/08 00:18:19 [lang:3] API nextsong - Response (522):                                                                                            
                           

2024/09/08 00:25:03 [lang:3] API nextsong - Response (522):                                                                                             ```

When you review app logs around these times, all of the sync tasks are fighting over locks.

Thank you.

BusterNeece:
@caturria We don't need to discuss that here. That should be a new issue. I'm fine with bumping the timeout as a solution for now, but let's discuss it somewhere else.

nunotfc:
Hello, One thing I would like you to check—I'm not opening a new issue because I can't prove it—is the repeat filter. My playlists don’t have repeated songs, and the filter is set for a 24-hour interval. However, the same song played again two hours after it had already played.
I believe this happened because the playlist was shuffled and finished playing. I don’t know how the repeat filter works internally, but perhaps each song should have a set time restriction, like "it can only play at a specific hour," and this should be enforced regardless of the playlist being shuffled, recreated, or running out of songs. Only clearing the play queue should reset the repeat filter.
Additionally, an option to retain the filter after a system restart could be added. As I mentioned, I didn’t open an issue because when it happened, I was heading out and forgot to capture the log.
I would also like the ability to set an interval longer than 24 hours. For example, if I have a playlist that lasts 3 days, I would prefer each song to play only once until the playlist is complete, even if the playlist is shuffled or randomized.
I have the impression that I keep hearing the same songs repeatedly, while many others are left unplayed. Of course, I’m not monitoring the radio 24/7, but whenever I do listen, it’s always the same songs, and there are over 3000 in the playlist. It might be just a confusion of me but, is thera a oportunity to check it when some dev is available to? thanks!

nunotfc:
"I agree, but I have no problem with repeating artists. I’m referring to playing the exact same tracks or song. For example, playing various songs by Rihanna within one hour is normal for me; playing the exact same song twice within two hours is not. I think your idea of separating artists and preventing duplicate titles will help a lot to fix this issue. Another thing that I think causes duplicates is the addition of songs. If I add new songs and the playlist is shuffled, they start playing immediately instead of waiting for the next shuffle or being played at the end of the currently shuffled playlist." but, if you want, contact me for maintaining this conversation. i think this place is not the right place to:

Issue #7426, opened 2024-09-03
Live Streamer/DJ Connected Webhook not sending during DJ Transition


Bloofield:
### Installation Method

Docker Installation

### AzuraCast Release Channel

Stable Channel

### Current AzuraCast Version

0.20.2

### What happened?

1. Have DJ 1 already on air
2. DJ 2 connects at the exact moment DJ 1 disconnects
3. No webhook sent for DJ 2 coming on air

PR #7427, opened 2024-09-03
Added support for DJ Transitions

Bloofield:
**Fixes issue:**
#7426 

**Proposed changes:**
- Adds support for when two DJs transition seamlessly.


t3duk:
Seems like this has caused some unexpected behaviour for me per chance.

PR #7447, opened 2024-09-16
feat: optimize Dockerfile

duhow:
**Proposed changes:**
This is a refactor of the Dockerfile.

- `strip` go binaries to remove unneeded debug symbols (saves ~30MB)
- move `COPY` into a single layer run `dependencies`
- Use `--mount bind` to avoid copying the build scripts into the Docker layers 
- Move `docs` copy to the end (in case there is a change in docs, the packages build will be rebuilt)

Although my main goal is to try squeeze the image size (currently about 1.8GB), I haven't played enough to see what else may be candidate for cleanup...

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,16 +32,9 @@ FROM ghcr.io/azuracast/icecast-kh-ac:2024-02-13 AS icecast
 #
 FROM mlocati/php-extension-installer AS php-extension-installer
 
-#
-# Final build image
-#
-FROM php:8.3-fpm-bookworm AS pre-final
+# ---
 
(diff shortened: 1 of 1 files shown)


BusterNeece:
@duhow Thank you for your contribution! I actually wasn't sure if the bind-mounted RUN command would work well on Depot, our CI build runner, but I tested it locally and it works just fine without any problems, so I'm happy to merge this to get that much-needed improvement in efficiency.

BusterNeece:
Seems like it shaved about 20MB off the compressed size on the Docker Hub. It's definitely something, and it was basically free to implement. Nice!

PR #7451, opened 2024-09-17
Make SFTP Connection URL hint the server URL, not the browser URL

yasiupl:
**Proposed changes:**
* Make the URL hint for SFTP connections [the same as Streamer URL](https://github.com/yasiupl/AzuraCast/blob/e38fa383b58a2cb9a739f8ccdf32008501e6c662/frontend/components/Stations/Streamers/ConnectionInfo.vue#L24C30-L24C54).
* I believe `serverURL` is the Azuracast prefered domain name, regardles of what domain the user uses in the browser. This is the domain name that will expose the actual server hosting Azuracast. The browser URL is routed via web proxies that might block SFTP/SSH (Cloudflare for example).  Thus, `serverURL` is the right choice here.

### DIFF
--- a/frontend/components/Stations/SftpUsers.vue
+++ b/frontend/components/Stations/SftpUsers.vue
@@ -50,7 +50,7 @@
                         <dt class="mb-1">
                             {{ $gettext('Server:') }}
                         </dt>
-                        <dd><code>{{ connectionInfo.url }}</code></dd>
+                        <dd><code>{{ connectionInfo.serverUrl }}</code></dd>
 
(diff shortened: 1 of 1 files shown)


BusterNeece:
Sorry for the delay on this. This is useful for us to do, but this isn't the way to do it unfortunately; the API being called by the "SFTP Users" page has different return values than the one used by the Streamers page, so it has to be changed on the PHP side. The latest commit to the Rolling Release achieves the same changes but from the server-side.

Issue #7681, opened 2025-01-14
Listener requests disrupting `Once per x Songs` playlist logic


RM-FM:
### Installation Method

Docker Installation

### AzuraCast Release Channel

Rolling Release Channel

### Current AzuraCast Version

Rolling Release #f0a9d37 (2025-01-02 20:42) 

### What happened?

It seems that requests are not being considered when counting the songs played before adding a track from a playlist that uses the `Once per x Songs` logic. Please see the attached screenshot that highlights the issue.

It's fine to exclude tracks assigned to Jingle Mode playlist from counting, as shown with the second-to-last item in the screenshot. However, the "Listener Request" item is a regular song and should not interfere with the `Once per x Songs` counter. I’ve even noticed worse situations where 2–3 requests were inserted, significantly delaying the jingle.

**Why is this critical to us?**
These jingles also serve as triggers for advertising slots. If we fail to play them, we risk missing opportunities to provide those slots to the ad injection system.

Not sure if this issue was hidden and came to our attention just now. Possibly it's affecting the Rolling Release only.

![image](https://github.com/user-attachments/assets/4cf58c68-df59-4e9b-86b8-8b6eb0dde752)
![image](https://github.com/user-attachments/assets/bc004dd2-cb71-4567-b84c-721eb3c4807a)

RM-FM:
The issue can be easily fixed in the `isPlaylistRecentlyPlayed()` function within `StationQueueRepository.php`. Currently, the condition `sq.playlist_id IS NOT NULL` prevents listener requests from being counted. 

```
public function isPlaylistRecentlyPlayed(
    StationPlaylist $playlist,
    ?int $playPerSongs = null
): bool {
    $playPerSongs ??= $playlist->getPlayPerSongs();

    $recentPlayedQuery = $this->em->createQuery(
        <<<'DQL'
            SELECT sq.playlist_id
            FROM App\Entity\StationQueue sq
            WHERE sq.station = :station
            AND sq.playlist_id IS NOT NULL
            AND (sq.playlist = :playlist OR sq.is_visible = 1)
            ORDER BY sq.id DESC
        DQL
    )->setParameters([
        'station' => $playlist->getStation(),
        'playlist' => $playlist,
    ])->setMaxResults($playPerSongs);

    $recentPlayedPlaylists = $recentPlayedQuery->getSingleColumnResult();
    return in_array($playlist->getIdRequired(), (array)$recentPlayedPlaylists, true);
}
```

[...]

PR #7682, opened 2025-01-15
Ensure stickyness of OncePerXSongs playlists in Jingle mode

RM-FM:
**Fixes issue:**
https://github.com/AzuraCast/AzuraCast/issues/7681

**Proposed changes:**
1. Ensure listener requests are taken into account
`backend/src/Entity/Repository/StationQueueRepository.php` > `isPlaylistRecentlyPlayed`
2. Give priority to `OncePerXSongs` playlists in `Jingle Mode` over listener requests
`backend/src/Radio/AutoDJ/QueueBuilder.php` > `getNextSongFromRequests`


RM-FM:
@BusterNeece @Vaalyn What do you think about this? Could we integrate it into the RR? We might also consider enhancing the playlist by incorporating a "prioritize over listener requests" flag as an alternative, rather than relying solely on "jingle mode" or the "once per x songs" rule. However, I don’t feel confident in my ability to implement the approach with an additional flag.

BusterNeece:
Hmm, I'm not sure I like this approach of just skipping the queue in such a specific hard-coded way. I'd rather we be able to have some more robust system of indicating when a track can be "jumped in front of" by requests, perhaps even storing that as metadata alongside the queued item.

RM-FM:
> Hmm, I'm not sure I like this approach of just skipping the queue in such a specific hard-coded way. I'd rather we be able to have some more robust system of indicating when a track can be "jumped in front of" by requests, perhaps even storing that as metadata alongside the queued item.

Certainly, that would make a lot of sense. Perhaps you could look into it at some point? I would be happy to see the hardcoded approach in the RR at least temporarily and refine it afterward - but it’s up to you…

PR #7700, opened 2025-01-24
Update Logs.vue

wegGehamstert:
The link on the log page redirect to a local not working docs page so it should forward to the main docs page

**Fixes issue:**

**Proposed changes:**

### DIFF
--- a/frontend/components/Stations/Logs.vue
+++ b/frontend/components/Stations/Logs.vue
@@ -43,7 +43,7 @@
                     </p>
                     <p class="card-text">
                         <a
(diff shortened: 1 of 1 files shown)


wegGehamstert:
It's of course a option to either fix the offline documentation or remove it completly.

BusterNeece:
Yeah, the actual issue here was that the local docs weren't rendering correctly. This was due to an upstream build issue; we do indeed want to keep the local docs! They're handy for users to have an exact offline copy of the docs for the version they're using.

The offline docs are fixed in the latest Rolling, which in turn fixes this link.

PR #7706, opened 2025-01-26
Update to PHP 8.4

BusterNeece:
This is currently depending on PHP-DI to tag a release version with PHP 8.4 compatibility, as otherwise its built container cache doesn't work correctly.

Otherwise, upstream libraries appear to be good with the exception of `E_DEPRECATED` errors, which are suppressed as they're outside our control in this case.

### DIFF
--- a/.github/workflows/default.yml
+++ b/.github/workflows/default.yml
@@ -37,7 +37,7 @@ jobs:
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
         with:
-          php-version: '8.3'
+          php-version: '8.4'
           extensions: intl, xdebug
           tools: composer:v2, cs2pr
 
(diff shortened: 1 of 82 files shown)


t0mtaylor:
Nice work @BusterNeece - i'm trying some optimisations to the www.conf and php.ini for opcache - I can push a PR up later this week - they will work on 8.3 as well as 8.4

Sella-GH:
Seems like PHP-DI tagged a 8.4 compatible version: https://github.com/PHP-DI/PHP-DI/releases/tag/7.0.8

BusterNeece:
@Sella-GH Excellent. In local testing everything's working fine. Merging into Rolling.

Sella-GH:
Update went through, looks promising! 🎉

Issue #5402, opened 2022-05-22
Error response from daemon: unknown log opt 'max-size' for journald log driver


pbu80:
### Installation Method

Docker Installation

### AzuraCast Release Channel

Stable Channel

### What happened?

tried to follow the unattended installation 

```
root@prabhu-desktop:/var# mkdir -p /var/azuracast
root@prabhu-desktop:/var# cd /var/azuracast
root@prabhu-desktop:/var/azuracast# curl -fsSL https://raw.githubusercontent.com/AzuraCast/AzuraCast/main/docker.sh > docker.sh
root@prabhu-desktop:/var/azuracast# chmod a+x docker.sh
root@prabhu-desktop:/var/azuracast# yes 'Y' | ./docker.sh setup-release
Your current release channel is 'Rolling Release'. Switch to 'Stable' release channel? [y/N] You should update the Docker Utility Script after changing release channels. Automatically update it now? [Y/n] New Docker utility script downloaded.
root@prabhu-desktop:/var/azuracast# yes '' | ./docker.sh install
Checking installation requirements for AzuraCast...
[PASS] Operating System: Linux
[PASS] Architecture: x86_64
[PASS] Command Present: curl
[PASS] Command Present: awk
[PASS] User Permissions
[PASS] Installation Directory
[PASS] All requirements met!
Docker is already installed! Continuing...
Docker Compose is already installed. Continuing...
[+] Running 1/1lease channel is 'Stable'. Switch to 'Rolling Release' release channel? [y/N]
 ⠿ installer Pulled                                                                                                                                                                     1.3s
the input device is not a TTY
[+] Running 1/1
 ⠿ web Pulled                                                                                                                                                                           1.2s
the input device is not a TTY
[+] Running 0/0
 ⠋ Container azuracast  Creating                                                                                                                                                        0.0s
Error response from daemon: unknown log opt 'max-size' for journald log driver
```

SC2Mitch:
Not entirely sure this is an AzuraCast problem as the error relates to Docker itself. 

Have you tried the other default installation method?

Vaalyn:
This as a problem with a setting of the Docker daemon itself.

Docker supports multiple ways to save container logs (those are called logging drivers). The default way in most cases is the `JSON File` logging driver. For this one the AzuraCast compose file definition contains a block to specify how big a single log file is allowed to be in order to prevent the logs from filling up the servers HDD.

[...]

PR #7710, opened 2025-01-29
Specify logging driver in docker-compose.sample.yml

nikolateslax:
This single-line addition specifies that the "json-file" driver is relied upon for AzuraCast. Since the script downloads and modifies this file, this should help address #5402.

**Fixes issue:** #5402 

**Proposed changes:**
- Specifies the "json-file" logging driver in the base Docker compose configuration, which is modified by the installer script.

### DIFF
--- a/docker-compose.sample.yml
+++ b/docker-compose.sample.yml
@@ -191,6 +191,7 @@ services:
         soft: 65536
         hard: 65536
     logging:
+      driver: "json-file"
       options:
         max-size: "1m"
         max-file: "5"


BusterNeece:
I'm worried that any time we influence the host installation's Docker settings, we'll have unexpected results. A vast majority of our users don't particularly have any interest in which logging driver is used for Docker, but in hard-coding it into our Docker Compose, we'd be messing things up for users who _do_ have reasons to use other drivers, in a way that they can't really negate via their own override file.

PR #7713, opened 2025-02-03
Feature: Add High-Precision DateTimes to Database Values

BusterNeece:
This PR represents an initial set of changes to improve the precision of datetime and duration values stored in AzuraCast. This not only serves to "future-proof" the application against possible future use cases, but allows us to have full microsecond resolution of track durations and expected playback, greatly improving the granularity of near-or-sub-one-second tracks (i.e. station ID bumpers) and avoiding overlap issues.

### Schema Changes

DateTime values in the following tables have been converted to MariaDB `DATETIME(6)` values, with microsecond precision:

- `AuditLog`
- `Listener`
- `Relay`
- `SongHistory`
- `StationQueue`
- `StationRequest`
- `StationPlaylist`
- `StationStreamerBroadcast`

Some of these don't directly relate to the goal of improving the AutoDJ scheduler, but could nevertheless benefit from improved time detail, so they were included to make the migration one big simple one.

Internally to the database, all datetimes are stored in UTC, in keeping with the prior UNIX timestamps. They're converted to local station time at the time of presentation. These DateTime values are passed around internally using the Carbon library, as it adds very useful functionality on top of DateTime; specifically, the values are `CarbonImmutable`, so as to avoid errors with manipulating dates down the line.

In addition to this, duration values in `SongHistory`, `StationMedia` and `StationQueue` have all been updated to be double-precision floating point values.

Relevant code in the AutoDJ scheduler has already been updated to take advantage of these new, higher-definition values when building the queue.

### API Changes

Public-facing APIs convert the internal `CarbonImmutable` values into UNIX timestamps in seconds for back-compatibility. Internal APIs have switched to using the ISO-8601 (`YYYY-mm-dd hh:mm:ss.uu`) format. When submitting values to be saved to the database, both UNIX timestamps and ISO-8601 dates are supported.

Tagging @caturria!


caturria:
Hi @BusterNeece,

First and foremost, I just wanted to express my admiration for how strong of mind you must be to be able to focus on this while the current unfathomable sequence of events play out in the White House. I'm not even American, I have an exam to study for, and all I can think about is Musk and his apparent world domination attempt from the White House.

Anyway, I might be able to jump back into this later in the week. It'll be very cool to start refining scheduling/ playout to take advantage of the improved precision.

I don't currently have an instance up at all or I'd be able to answer this myself, but is the Autocue binary currently available in the image?

If not, as I'm currently quite wet behind the ears when it comes to the Docker side of things, might you be cool with installing a copy?

Looking forward to working with you again,

Jordan (Caturria).

BusterNeece:
@caturria Trust me, it's been a stressful few weeks. Honestly, the work on AzuraCast is a welcome distraction from it all. Code is something I know and can influence for the better. :)

I'm happy to work with you on installing things, just let me know. Currently, the AutoCue binary isn't present in the Docker image as we removed it after it was no longer going to be maintained.

caturria:
Oh, Wow... I just read up on @Moonbase59's situation. If you see this, I'm sorry. God, the best thing that could possibly happen at this point is if the world just ended somehow.

As far as moving forward goes, I'm assuming Liquidsoap is just maintaining their own integration at this point?

BusterNeece:
@caturria Their own internal Liquidsoap-based implementation, yes, but not a standalone binary. We could still invoke it via a custom small Liquidsoap script that phoned back home to us, but it would be a little more error-prone than just executing an AutoCue binary ourselves.

PR #7832, opened 2025-04-03
Prioritize playlist over listener requests

RM-FM:
This adds an additional checkbox in the advanced playlist options to allow certain playlists to play before listener requests.

**Fixes issue:**
(https://github.com/AzuraCast/AzuraCast/issues/7681)


RM-FM:
> @RM-FM Very good, thank you. I will take a look at this as soon as I have time.

Hi @BusterNeece, did you already manage to have a look at it? If you need some more time, would it be possible to create a dedicated feature branch?

RM-FM:
@BusterNeece I just wanted to kindly follow up on the code I shared - no pressure at all, just checking in. If a merge into main isn’t possible yet, would it be okay to set up a feature branch in the meantime? I've noticed that Jingles often play much later than expected (after 7–8 songs instead of 3), when multiple requests come in. Since the Jingles trigger ad slots, this delay also affects our revenue.

RM-FM:
@BusterNeece Thaaaaank you! :-)

PR #7834, opened 2025-04-05
Update Crowdin configuration file

gInAmOsLaMi:
**Fixes issue:**

**Proposed changes:**


Vaalyn:
🤔 Crowdin's own docs about the conf file says the param `preserve_hierarchy` should be marked with a boolean, not an integer. Why would this change be neccessary? 

https://support.crowdin.com/developer/configuration-file/

BusterNeece:
I don't think it is necessary. Closing this for now.

PR #7962, opened 2025-07-06
refactor: property descriptions in StationMedia for clarity and consistency

mdrxy:
If I understand correctly,
* `song_id` is a hash that identifies the actual audio file. If you were to upload the same song twice, both files will have the same `song_id`, useful for detecting duplicate media files.
* `unique_id` is for the specific database record of a media file in a station's library. Each entry in the media table gets its own `unique_id`, even if it points to a song that already exists elsewhere (sharing a `song_id`).

This wasn't initially clear in the API docs, so I propose these updates to help clarify things for future users.

### DIFF
--- a/backend/src/Entity/Api/StationMedia.php
+++ b/backend/src/Entity/Api/StationMedia.php
@@ -21,26 +21,26 @@ class StationMedia
     use HasLinks;
 
     #[OA\Property(
-        description: 'The media\'s identifier.',
+        description: "The media's identifier.",
         example: 1
     )]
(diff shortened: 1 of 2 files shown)


BusterNeece:
@mdrxy Slight clarification in that the song ID is based on the _metadata_ of the song (i.e. a hash of title and artist), not the actual _contents_ of the song. It's not an acoustic hash i.e. something Shazam would use. Would be kinda cool if we could pull something like that off, but nope.

PR #8016, opened 2025-07-23
feat: add bulk media preview functionality

mdrxy:
Introduce a new endpoint for previewing changes before importing bulk media files. This feature allows users to see proposed changes and confirm them prior to the actual import process.

Screenshot:
Screenshot


BusterNeece:
@mdrxy Since it overlaps so heavily with `UploadAction`, I just made it a "sub-action" of that to avoid duplicating a lot of code for both, but otherwise it looks just fine to me. Merged!

mdrxy:
@BusterNeece thank you! The next idea I had (but struggled to implement) was for the modal that appears after an upload (the one that shows the status)

Currently it shows a status for each media row submitted, which can be redundant if only a small subset of media items actually changed. E.g. if you're uploading a `.csv` with 300 media items, 300 status messages is shown, even if you only modified 5 of the 300. Does that make sense? The remaining 295 could be collapsed into something along the lines of "295 media items processed but unchanged"

PR #8028, opened 2025-07-29
performance improvements and segfault protection

t0mtaylor:
**Fixes issue:**
https://github.com/AzuraCast/AzuraCast/issues/7988 

**Proposed changes:**
Performance tweaks to improve hosting of Azuracast

Also allows application to startup without segfaults


BusterNeece:
I pulled a few of these items over, but a few override settings that are defined in other files (i.e. FPM's conf overriding php.ini) and some I think are probably too aggressive for the standard installation.

t0mtaylor:
Great thanks @BusterNeece

PR #8073, opened 2025-08-17
Update LS 2.4.0-RC1

gAlleb:
Memory leak fix

**Fixes issue:**

**Proposed changes:**


gAlleb:
However, wait a minute. I think some changes should be made regarding synchronous threads, insert_metadata and so on. https://www.liquidsoap.info/blog/2025-08-11-liquidsoap-2.4.0/

PR #8125, opened 2025-09-13
feat(broadcast): Simulcasting

Yahav:
This introduces a new feature: Simulcasting, eg. broadcasting simultaneously to various providers (currently Facebook and YouTube).
Inspired by https://github.com/AzuraCast/radio-video-stream
There are plenty of improvements that could be made but at this point its working at the basic level.

My main problem with this feature is that LS doesn't provide true observability for the actual connection which means the status of the instances might not reflect the actual status of the connection. (https://github.com/savonet/liquidsoap/issues/4672)

**Important:**

- This should be code reviewed and thoroughly tested
- This was based on the stable branch and was created for LS 2.3.3, not sure how merging into main will work
- It was tested on my Ansible build so you should probably test the Docker installation. I've added the basic required modifications to the docker assets but not quite sure how to test everything is working.
- I wasn't sure how to add the video.mp4 file properly, its about 10mb which makes the main repo too big imho, perhaps uploading those assets to a separate repo and downloading them at build?

### DIFF
--- a/frontend/entities/ApiInterfaces.ts
+++ b/frontend/entities/ApiInterfaces.ts
@@ -1765,6 +1765,21 @@ export type SftpUser = HasAutoIncrementId & {
   publicKeys?: string | null;
 };
 
+export type Simulcasting = HasAutoIncrementId & {
+  /** @example "My Facebook Stream" */
+  name?: string;
+  /** @format int64 */
+  station?: number;
+  /** @example "facebook" */
+  adapter?: string;
+  /** @example "your_stream_key_here" */
+  stream_key?: string;
+  /** @example "stopped" */
+  status?: string;
+  /** @example "Connection failed" */
+  error_message?: string | null;
+};
+
 export type Station = HasAutoIncrementId & {
   /**
    * The full display name of the station.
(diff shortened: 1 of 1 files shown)


BusterNeece:
@Yahav Thank you for the submission! This is quite a lot of work, and honestly it looks quite promising, but I think it needs something of an overhaul to really integrate cleanly with the software.

Few things of note:

- Since the PR's submission and now, we've made some big structural changes (making Doctrine entities use public properties and property hooks, using Regle for forms on the frontend, etc.) and it would be preferable if this PR was updated to those things.
- I'm not sure that Simulcasting benefits from having a standalone channel in Centrifugo, rather than, say, a separate message sent  on its own channel, which feeds into the next point:
- Even though there's support for multiple simulcasting streams in multiple stations at the same time, some of the configuration seems like it's built for a single simulcast stream to be running on an installation at a given moment. Not sure if that's the case?
- I'm not actually sure "Simulcasting" is really the best word for this. I prefer language, if possible, that represents what something does in simple, easily translated words, hence why the repo for the radio-video-stream was called that.

In its current state, it just needs too much work for me to really be able to spend the time needed to get it merged in. If you really want to see the feature ship in upstream AzuraCast, you're welcome to either update this PR or resubmit it as a new one and I'm happy to re-review it.

PR #8130, opened 2025-09-14
chore: upgrade Regle and patched some DX

victorgarciaesgi:
Hi!

I'm the maintainer of the library [Regle](https://github.com/victorgarciaesgi/regle)

I've came across this project when I saw you were using my library and I'm so happy a project this big is using it 👌 

## Changes

I've allowed myself to improve a little bit of the experience and usage, even if this project make a great use of some advanced features.

- Upgrade Regle to latest version which includes some bug fixes
- Fixed the typing problem you had with validation groups and variants
- Removed occurence of `useResettableRef` for form, as Regle provides `$reset({toOriginalState: true})` that does the same thing. (Did not updated all the occurrences yet)
- Improved global config do avoid double rule message declaration

Don't hesitate to ask me if there is any feature that would be useful to you, seing it used on big project like this helps me a lot to improve it 👍 

## Problems

I couldn't test locally because of a Docker environnement issue, one container (Container azuracast) fails to start because of this error

```ts
(HTTP code 500) server error - driver failed programming external connectivity on endpoint azuracast (cde359f41497ebf458495688ba49929005178dcb177af1ea7d6d575a8f108): Bind for 0.0.0.0:80 failed: port is already allocated
```

I tried killing the port and still same error


BusterNeece:
@victorgarciaesgi Hey!! This is super cool. It's quite rare for me to see the maintainer of a package we use here come and help out with its use directly, and I think that's awesome.

First of all, you should know that Regle is _fantastic_. It has solved so many problems for us and been a natural successor to Vuelidate. Without it, I highly doubt we could be in "strict mode" on TypeScript nowadays, which is something I've wished we could do for years.

Secondly, I actually had no idea about the `toOriginalState`...I only saw the `toInitialState` and thought perhaps it wouldn't work for me because of the caveats it had, so I just stuck with my existing little composable. Seeing that such a thing exists, I'll go ahead and swap over every remaining item to use it.

Also...you removed that TypeScript error notation I had on the variants, but in my local tests, that actually caused the TS error to pop right back up.

It looks like:

```
error TS2322: Type '(fields: { readonly path?: RegleFieldStatus<string | undefined, {}, {}> | undefined; readonly type?: RegleFieldStatus<StorageLocationTypes | undefined, {}, {}> | undefined; ... 21 more ...; readonly storageAvailable?: RegleFieldStatus<...> | undefined; } & { ...; }) => { ...; }' is not assignable to type '(fields: { readonly path?: RegleFieldStatus<string | undefined, {}, {}> | undefined; readonly type?: RegleFieldStatus<StorageLocationTypes | undefined, {}, {}> | undefined; ... 21 more ...; readonly storageAvailable?: RegleFieldStatus<...> | undefined; } & { ...; }) => Record<...>'.
  Type '{ generalTab: (RegleFieldStatus<string | null | undefined, {}, {}> | RegleFieldStatus<StorageLocationAdapters | undefined, { literal: RegleRuleDefinition<StorageLocationAdapters.Dropbox, ... 4 more ..., string | number>; required: RegleRuleDefinition<...>; }, {}> | RegleFieldStatus<...> | RegleFieldStatus<...> | Reg...' is not assignable to type 'Record<string, RegleValidationGroupEntry[]>'.
    Property 'generalTab' is incompatible with index signature.
      Type '(RegleFieldStatus<string | null | undefined, {}, {}> | RegleFieldStatus<StorageLocationAdapters | undefined, { literal: RegleRuleDefinition<StorageLocationAdapters.Dropbox, ... 4 more ..., string | number>; required: RegleRuleDefinition<...>; }, {}> | RegleFieldStatus<...> | RegleFieldStatus<...> | RegleFieldStatus<...' is not assignable to type 'RegleValidationGroupEntry[]'.
        Type 'RegleFieldStatus<string | null | undefined, {}, {}> | RegleFieldStatus<StorageLocationAdapters | undefined, { literal: RegleRuleDefinition<StorageLocationAdapters.Dropbox, ... 4 more ..., string | number>; required: RegleRuleDefinition<...>; }, {}> | RegleFieldStatus<...> | RegleFieldStatus<...> | RegleFieldStatus<....' is not assignable to type 'RegleValidationGroupEntry'.
          Type 'undefined' is not assignable to type 'RegleValidationGroupEntry'.
            Type 'undefined' is not assignable to type 'Omit<RegleCommonStatus<any>, "$value" | "$silentValue">'.

74                 validationGroups: (fields) => ({
                   ~~~~~~~~~~~~~~~~


Found 1 error in frontend/components/Admin/StorageLocations/Form/form.ts:74
```

I think the TS validator really doesn't like the whole variants thing and the possibly-undefined fields as they apply to validation groups. They seem to work perfectly in practice, though, so I just told the TS compiler to ignore it for now. It may be worth investigating on the library side, though!

victorgarciaesgi:
Thank you a lot for the kind words! Super happy to keep the Vuelidate way to do forms alive 💪 

1. Sorry I haven't been obvious in my message, `toOriginalState` is actually a new feature from [v1.8.0](https://github.com/victorgarciaesgi/regle/releases/tag/v1.8.0) so that's normal if you missed it 😄 

The difference from `toInitialState` is that it's never modified, and always keep the first value you passed as state, whereas initialState can be patched when reseting validation

2. The Typescript error was a problem on Regle's side that I discovered because of your issue!

I patched it in the latest version [v1.8.4](https://github.com/victorgarciaesgi/regle/releases/tag/v1.8.4), and that's why I could remove the comment on your codebase 👌 

3. Does the error I have while trying to reproduce the local env sounds a bell? I'm on M1 Pro mac running Sequoia 15.6.1

victorgarciaesgi:
Ok found out I had other docker container started from another project that made conflicts with ports 👍

BusterNeece:
@victorgarciaesgi Hey, update! I went through and updated a lot of stuff in the code to use Regle's built-in features for things like resetting form state, and I honestly think I created more TypeScript errors than I fixed heh.

In particular, I think I found an issue where Regle is "losing" the required state for a property if it doesn't literally have the `{required}` rule. In other words, if you pass:

```typescript
type Foo = {
  a: string,
  b: boolean
}

const form: Foo = {
  a: '',
  b: false
}

const {r$} = useRegle({
  form,
  {},
  {}
});

const {valid, data} = await r$.$validate();
// The type of "data" will now be, more or less:
// { a?: string | undefined, b?: boolean | undefined }
```

Since I'm now using that return value instead of just directly hitting the "parent" ref, that typing difference is throwing me off in a few spaces. Any thoughts on that?

victorgarciaesgi:
@BusterNeece Nice!

That's not a bug but on purpose. The trick is: 
- Even if your `Form` interface does not marks `a` and `b` as optional, Regle will assume the fact that you're working on a form. And in a form, every value can be modified or cleared. Type wise and staticaly, it has no clue of what would happen.
- The only "guard"  is the `required` rule, because with this rule declared, it ensure the value will be present 100% of the time after validation. Otherwise it could be set to `undefined` by any means

PR #8131, opened 2025-09-14
feat(auth): SSO Login Url

Yahav:
This allows an admin to create a single time use, time limited login url for a specific user.
This would be very handy in situations such as redirecting a user from a customer portal systems like WHMCS.
Onces merged i will implement the SSO feature at the WHMCS plugin:
https://github.com/Yahav/whmcs-azuracast


BusterNeece:
@Yahav We already have the ability to generate temporary login tokens via CLI...why not just use that system as it exists in the app instead of building what is basically a copy of it?

I can look into tying the two together if you'd like...

Yahav:
@BusterNeece I've looked into the current reset token mechanism and it seem like it doesn't actually connect you to the app but just allow access to the password recovery page and so i went ahead with this new mechanism.

>I can look into tying the two together if you'd like...

But of course, if you see a better way to achieve this, by all means :)

BusterNeece:
@Yahav I opted to take a somewhat simpler approach and just expand the `UserLoginTokens` to support both types and optional messages/expiration dates. That offers the ability to create "magic links" via the `POST /admin/login_tokens` endpoint without overly complicating the existing process, while still supporting the existing password reset function.

PR #8173, opened 2025-10-02
Fix Issue #8172 - Use composable to return userAllowedForStation function to fix composable usage outside of setup when calling useStationId

### LINKED ISSUE #8172: Sidebar Menu Not Displaying After Reload/Restart in AzuraCast v0.23.1
### Installation Method

Docker Installation

### AzuraCast Release Channel

Stable Channel

### Current AzuraCast Version

 v0.23.1

### What happened?

After changing the Liquidsoap configuration in AzuraCast v0.23.1 (Stable), the system prompts to reload the configurations to apply changes. However, when I attempt to reload or restart, the sidebar menu does not fully load. Only the "View Profile" option is shown, and the rest of the sidebar menu is missing.

To resolve the issue, I have to manually refresh the page in the browser, which causes the sidebar to load correctly.

Steps to Reproduce:

Change the Liquidsoap configuration in the AzuraCast interface.

Reload or restart the system to apply changes.

Observe that only the "View Profile" option is visible in the sidebar.

Refresh the page manually, and the sidebar will load fully.

Expected Behavior:
The sidebar should fully load after a reload or restart without needing a manual browser refresh.

Environment:

Browsers: (Chrome, Firefox)

Possible Fix:
It seems like the sidebar might not be properly re-rendering after a configuration reload, which may be related to how the frontend is being updated in the backend.

### Relevant log output

```shell

```

Vaalyn:
**Fixes issue:**
Fixes #8172 

**Proposed changes:**
This PR moves the `userAllowedForStation` into a composable function that returns the `userAllowedForStation` function in order to fix the call to `useStationId` causing an error to be thrown when it is used outside of the initial setup rendering pass of the UI cycle which causes the `visible()` calls for the menu items to fail as `false` in the catch.

This is what I got in the browser console when investigating this problem:
```
[Vue warn]: inject() can only be used inside setup() or functional components.
```

This is the error that is being `catch`'ed when this happens:
```
TypeError: (intermediate value)() is undefined
useStationId useStationQuery.ts:10
userAllowedForStation acl.ts:22
...
```

### DIFF
--- a/frontend/acl.ts
+++ b/frontend/acl.ts
@@ -16,37 +16,33 @@ export function userAllowed(permission: GlobalPermissions): boolean {
     }
 }
 
-export function userAllowedForStation(permission: StationPermissions, id: number | null = null): boolean {
-    if (id === null) {
-        try {
-            const stationId = useStationId();
-            id = stationId.value;
-        } catch {
-            return false;
+export function useUserAllowedForStation() {
+    const {permissions} = useAzuraCastUser();
+    const stationId = useStationId();
+
+    return {
+        userAllowedForStation: (permission: StationPermissions) => {
+            try {
+                if (userAllowed(GlobalPermissions.Stations)) {
+                    return true;
+                }
+
+                const thisStationPermissions = permissions.station.find(
+                    (row) => row.id === stationId.value
+                );
+
+                if (thisStationPermissions === undefined) {
+                    return false;
+                }
+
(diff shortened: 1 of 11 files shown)


BusterNeece:
@Vaalyn Yeah that's definitely the right solution to this, though I regret now having the station perm composable be, well, a composable, while the global-level ACL isn't. Might just make it one despite not needing to be, just to be consistent.

Vaalyn:
👍 yeah, makes sense to keep it consistent. Didn't think of that when making the change 😅

PR #8178, opened 2025-10-05
UI/UX and Semantic Thoughts/Suggestions

Yahav:
This suggests two changes that really make sense to me.

### The first is a minor UI/UX restructure:

Currently AzuraCast station dashboard has what i think are three UI/UX issues:
1. the _rather important_ Station Start / Station reload alerts are placed at the sidebar which is too small and possibily easy to miss. 
2. There are 3 instances for the "Edit Profile" links which are redundant, same goes for the station name which isn't important enough to be displayed twice and in such good real estate.
image

So what i did to fix those is to move the station clock and station name to the main navbar header,
Get rid of the dashboard header altogether and move the play button to the header of the now playing card.,
move the station alerts to the main container header which you cannot miss even if you wanted to.

### The second is a minor semantic change regarding the Profile

I always thought that Profile is rather the wrong terminology for this use (granted, english is my second language)
"View Profile" is basically a dashboard/overview of the streaming services/status
and "Edit Profile" is essentially the configurations/settings of the entire operation (which includes a section dedicated to the actual station profile)
So i opted to call things what they seem to be.
"Overview"
and
"Settings" / "Station Settings"

Caster_fm-8

Once again, those are marely suggestions that to me **personally** makes more sense, it might be more suitble for Discussions rather than a PR but i wouldn't want to suggest something without actually seeing how it looks.
:)


BusterNeece:
@Yahav Hmm. I'm kinda split down the middle on this one.

Things I like:
 - Calling Profile "Overview", and Edit Profile "Station Settings" or just "Settings"
 - Reducing the number of redundant links to the edit-profile functionality

Things I probably wouldn't change:
 - Moving the station name out of the sidebar (the header is used for media playback, and on smaller devices is very narrow, while the sidebar is a fixed-width thing that opens up on smaller screens, so it's okay to put the station name/clock there)
 - Moving the "needs restart" alerts into the main body (because if they're there on every page, then they'll look slightly different depending on the page they're on, and if they're not, then users might miss that alert after making a change on a non-settings page)

There is _some_ method to the madness of having redundant links in various places that end up going to the same place, which is that as I watched users trying to use the software over the years, even though there was a link for what they were looking for somewhere else, they still found it faster and easier to have that link in a different spot as well.

Yahav:
Got it.
I'll return the station name and clock to the sidebar.

What about the removal of the dashboard header and placement of the play button besides the now playing header?

As for the stations alerts, i still think that in the sidebar they are easier to overlook.
In the current implementation i sent it will be displayed on every single station page.. 
>then they'll look slightly different depending on the page they're on

How so, perhaps we could hardcode some css to #station-time-wrapper. to take care of this..

BusterNeece:
@Yahav I've actually implemented much of these same changes to the latest Rolling Release, but via individual commits that address each separate component. I also polished up several other things in the overview/profile page, i.e. the badge layout and its legibility against the blue background.

renatopinto89:
I'm sorry for invading this already closed Pull Request, but would it be too ambitious to move the "Profile" Tab to  the "Branding" and only keep the technical stuff like Ports, Passwords, Auto-DJ features etc in the "Settings" section, so it's better manageable with permissions?
Like, not having the "Settings" permission but still allowed to change the stations name, URL stub, etc...

Best regards.

PR #8203, opened 2025-10-18
Player Embed Widget Builder

finnie2006:
This pull request introduces a  widget customization system for the public radio player, allowing appearance and behavior to be controlled via URL parameters and backend configuration. The changes span backend, frontend, and styling code, enabling features like custom colors, layouts, rounded corners, popup player, continuous playback, and more.  Should resolve this feature request https://features.azuracast.com/suggestions/39025/player-embed-widget-builder

**Widget Customization Framework**

* Added the `WidgetCustomization` class (`backend/src/Entity/Api/WidgetCustomization.php`) to parse and validate widget customization options from URL parameters, including colors, layout, feature toggles, and custom CSS.
* Integrated `WidgetCustomization` into the backend response (`NowPlayingComponent.php`), passing customization options to the frontend via the new `widgetCustomization` prop. [[1]](diffhunk://#diff-75a08dda5404755d1a79fb2dfcc06c2300cf42b597bd85f05e5ee81bb7eb6166R8) [[2]](diffhunk://#diff-75a08dda5404755d1a79fb2dfcc06c2300cf42b597bd85f05e5ee81bb7eb6166L21-R34)

**Frontend Integration and Feature Toggles**

* Updated `Player.vue` to use the new `widgetCustomization` prop, applying customization options for album art, autoplay, volume controls, track progress, stream selection, popup player button, and initial volume. [[1]](diffhunk://#diff-a67eb57195f388c3cccb542b14527887a827befa506e3c695d69abe68c21437dL2-R9) [[2]](diffhunk://#diff-a67eb57195f388c3cccb542b14527887a827befa506e3c695d69abe68c21437dL41-R45) [[3]](diffhunk://#diff-a67eb57195f388c3cccb542b14527887a827befa506e3c695d69abe68c21437dL73-R77) [[4]](diffhunk://#diff-a67eb57195f388c3cccb542b14527887a827befa506e3c695d69abe68c21437dL108-R125) [[5]](diffhunk://#diff-a67eb57195f388c3cccb542b14527887a827befa506e3c695d69abe68c21437dR164-R189) [[6]](diffhunk://#diff-a67eb57195f388c3cccb542b14527887a827befa506e3c695d69abe68c21437dL164-R215) [[7]](diffhunk://#diff-a67eb57195f388c3cccb542b14527887a827befa506e3c695d69abe68c21437dR230-R286)
* Implemented popup player functionality and continuous playback state persistence using localStorage, based on widget customization options. [[1]](diffhunk://#diff-a67eb57195f388c3cccb542b14527887a827befa506e3c695d69abe68c21437dR337-R434) [[2]](diffhunk://#diff-a67eb57195f388c3cccb542b14527887a827befa506e3c695d69abe68c21437dR481-R610)

**Styling and Layout Customization**

* Enhanced `Player.vue` styles to support custom CSS variables, layout variants (vertical, compact, large), and rounded corners, all controlled via customization options. Custom CSS from the backend can be injected dynamically. (F97a520dL478, [[1]](diffhunk://#diff-a67eb57195f388c3cccb542b14527887a827befa506e3c695d69abe68c21437dR678) [[2]](diffhunk://#diff-a67eb57195f388c3cccb542b14527887a827befa506e3c695d69abe68c21437dL373-R693) [[3]](diffhunk://#diff-a67eb57195f388c3cccb542b14527887a827befa506e3c695d69abe68c21437dR711-R718)

**Dependency Update**

* Added `@codemirror/commands` to `package.json` dependencies.

image

image

### DIFF
--- a/backend/src/Entity/Api/WidgetCustomization.php
+++ b/backend/src/Entity/Api/WidgetCustomization.php
@@ -62,7 +62,11 @@ final class WidgetCustomization
     #[OA\Property(description: 'Whether to persist playback state across pages.', example: false)]
     public bool $continuousPlay = false;
 
-    #[OA\Property(description: 'Additional CSS applied to the widget.', example: '.radio-player-widget { border-radius: 12px; }', nullable: true)]
+    #[OA\Property(
+        description: 'Additional CSS applied to the widget.',
+        example: '.radio-player-widget { border-radius: 12px; }',
+        nullable: true
+    )]
     public ?string $customCss = null;
 
     public static function fromRequest(ServerRequest $request): self


BusterNeece:
@finnie2006 Wow, a quite comprehensive feature here. There's a whole lot about this that I like and basically feel needs no extra modification, so I'll focus here on the things I either need clarity on or that we should revisit before merging.

- The biggest thing is that I _really_ don't like how the JS embed script is done. If we ever were to change the styling or structure of the HTML on the player (which we do quite a lot), it would require that we modify it in two very different places, one of which isn't even managed by our Vite build system. This is a pretty huge maintenance burden for what is, after all, not even the recommended way of embedding the player. I honestly think if someone's to the point of "power user" that they need an on-site embed instead of an iframe, then they can just directly use our code and API as a jumping-off point for a custom implementation (as many professional stations have) that they maintain themselves. I'd ditch this feature entirely and focus exclusively on the iframe side.

- I can't figure out what "Enable Popup Player" or "Continuous Play Across Pages" actually does. Can you clarify?

- If you have a `WidgetCustomization` API object inside PHP, you can sync that up with Typescript types by giving it the OpenAPI attributes (see other API objects) and then just running `npm run generate-api`, which updates the TS interfaces to match the PHP ones. Then you can just depend on that interface instead of having to define it again inside TS (and then you only have to actually update it in one place if it changes).

- What's `@codemirror/commands` used for in this context? We should already have it as a product of using Codemirror...

- The actual widget builder could use a polish pass to add some padding between elements (to avoid text butting up directly against controls above it), to add a "Reset" button to the color pickers to disable them if they're changed, and to line up the volume slider with its indicator. Given its new complexity, it might even make sense to split it into smaller little Vue components and pass around something like a Pinia store for all the settings (similar to how we do multi-tab forms) so we're not dealing with a super big component.

I'm happy to help with any or all of these if needed. Let me know if you're able to work on this or if you'd like some help with it. Once we get all these things figured out, it's absolutely in a merge-worthy state.

finnie2006:
**1.  JS embed script**

I do agree with your statements, i will just remove the js embed feature and for those kinds off power users, it is indeed just a better practice to use the api.

**2.  "Enable Popup Player" or "Continuous Play Across Pages"**
Enable player popup makes the player open as a popup instead off being visible directly on page load, and the feature continuous play across pages will make the player play when people navigate through the site.

**3. WidgetCustomization API object sync**
I did not think off that and indeed it seems like a good practice to implement.


**4. @codemirror/commands**

It is indeed already covered, so it can be removed.

**5. refinements**

I will work on that, and i will brief you how far i got on those points and if i need any help on this or things mentioned above.

finnie2006:
I think i implemented all your requests properly, but definitely you should check if i did them all correctly, and if the css is properly done!

BusterNeece:
@finnie2006 I went ahead and gave the whole thing a once-over and made sure everything matches our existing standards. Everything looks good now and it's merged in.

One big thing that was causing trouble is that the `Player.vue` component is also used inside `FullPlayer.vue`, so I had to split out the embed-specific stuff into a new `EmbedPlayer.vue` component to avoid wrecking how the full-page player looks.

Other than that, it was mostly just a matter of using vueUse's `useLocalStorage` instead of manually writing to it, and of using unhead's support for writing CSS/JS to the HTML header to more safely manipulate the header from within Vue.

finnie2006:
Yeah i just looked at your changes and indeed some better readability in the first case, and i completely made an oversight that the Player.vue component is used in the fullplayer, But overall i think a great new addition to AzuraCast!

StefCoders:
I'm having problems with this player, nothing in the console. Album art and metadata has loaded sucessfully but the play buttons and volume is frozen, clicking does nothing. The stream selector works but no activity when using buttons.

PR #8314, opened 2025-12-27
Switch to Upstream Icecast 2.5

BusterNeece:
This is still currently a work in progress pending additional development on the part of the Xiph Icecast team, but we'd like to get back to using the "upstream" Icecast (Icecast itself, not the now-largely-unmaintained KH branch) for AzuraCast.

This PR (along with changes to the build process that adds our custom HTML and MP3 files) switches us back to Icecast's latest commit version.

Some pending items that need addressing:
 - The `<x-forwarded-for>127.0.0.1</x-forwarded-for>` option does not exist in upstream Icecast yet, but is currently in the process of being reviewed for inclusion. It will be absolutely necessary for us to operate Icecast in the containerized environment we're in.
 - The `<deny-agents>` list, which KH let us supply as a single flatfile to avoid URL authentication for this scenario, doesn't exist in upstream Icecast, forcing us to go back to using URL-based authentication for all users if a user-agent deny-list is in place on the station's configuration.
 - For some reason, there appears to be a bug where Liquidsoap attempting to call `listclients` on a source that (until it later broadcasts to it) doesn't exist counts as one "source" broadcasting, so the actual connection attempt it makes hits the source limit; the quick workaround is to just make the source limit double what the mount point count is, but this is probably something to review with Xiph.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ FROM ghcr.io/azuracast/azuracast.com:builtin@sha256:f7c41278613d113375ca285ba85f
 #
 # Icecast-KH with AzuraCast customizations build step
 #
-FROM ghcr.io/azuracast/icecast-kh-ac:2025-08-29 AS icecast
+FROM ghcr.io/azuracast/icecast-ac:feat-switch-to-upstream AS icecast
 
 #
 # PHP Extension Installer build step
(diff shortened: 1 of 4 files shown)


Vaalyn:
Just did a quick check on how the blockers for this are coming along:
- The `x-forward-for` thingy seems to be merged https://gitlab.xiph.org/xiph/icecast-server/-/work_items/1959 ✅ 
- Deny User Agents set to mile stone 2.5.1 it seems: https://gitlab.xiph.org/xiph/icecast-server/-/work_items/2119 ⏳ 
- The "call listclients on a source" I haven't checked ❓ 

Should this PR be closed? The GH Actions bot closed it as stale a while ago.

PR #8340, opened 2026-01-22
Create docker-compose.yml

atknklkn:
**Fixes issue:**

**Proposed changes:**

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,219 @@
+#
+# AzuraCast Docker Compose Configuration File
+#
+# When updating, you will be prompted to replace this file with a new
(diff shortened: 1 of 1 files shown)


BusterNeece:
We deliberately don't commit a `docker-compose.yml` to our repo because development environments often want to reduce the number of ports that are defined in our Docker Compose config, and you can only _add_ configuration via `docker-compose.override.yml`, not _remove_ it, so to do that requires replacing the main file, which would be quite pesky to do if it was committed in the repo.

Our Docker utility script will automatically configure a `docker-compose.yml` file for you, but if you're doing local dev, you can just copy `docker-compose.sample.yml` to `docker-compose.yml` to get a working version.

PR #8433, opened 2026-04-09
Add playlist groups feature for grouping & nesting playlists

Vaalyn:
**Fixes issue:**
X

**Proposed changes:**
Work-In-Progess Branch for implementing a Playlist Groups feature that enables station operators to create nested playlists. (https://features.azuracast.com/suggestions/124222/clockwheels-a-very-important-feature-for-a-radio-station).

The implementation aims to integrate the logic for handling `Playlist Group` type playlists as gentle as possible into the current QueueBuilder logic so that the difference between the logic trees between regular and grouped playlists stays as minimal as possible.

I haven't yet fully tested everything through but the logic is working without running into errors and it is playing items from the grouped / nested playlists.

I'm planning to add a bunch of unit tests for the QueueBuilder in order to check that these things work as intended.

The UI allows drag & drop for assigning & ordering and is aimed to feeld like a directory structure.

## Features

- Playlist Groups
  - Introduces a new Playlist Source `Playlist Group` that allows defining groups of playlists
    - Allows nesting of playlist groups for better reusability and more fine grained controll over playback
  - Adds a drag & drop interface for managing playlist groups
  - Supports configuring an amount of consecutive plays for each member to enable users to create clockwheel's
    - Inspired by https://github.com/AzuraCast/AzuraCast/pull/8434 & https://github.com/AzuraCast/AzuraCast/discussions/8432
  - Supports configuring what kind of requests are allowed for each member to control what actually is pulled from request when the member is active
- Request Queue Playlists
  - Introduces a new Playlist Source `Request Queue` which enable more fine grained control over how listener requests should be scheduled / played
- Schedule Item option `Block Request Queue While Active`
  - Introduces a new schedule item option to allow blocking the playback of listener requests via the automatic 
request queue in the schedule window
  - This supersedes the toggle from the Advanced tab due to that one behaving in a non-intuitive way, blocking playback of all listener requests if set on a non-scheduled, general playback playlist
- Station wide option `Only play requests via Request Queue playlists`
  - Introduces a new option in `Station` -> `Song Requests` to hand over the control of the listener requests playback to `Request Queue Playlists`
  - Prevents playback of listener requests via the automatic request queue while keeping the request system active in general
- Playlist Configuration Exporter & Importer
  - Allows exporting either single playlists or all playlists of a station as a JSON dump
  - The JSON dump can be imported on other stations to re-create the playlists there while re-linking all associated media, folders & groups to them
  - Importer will generate dummy media files for any missing media while assigning them the value `DUMMY` as the ISRC so they are easy to identify
  - Mainly developed for easier testing of `QueueBuilder` & `Scheduler` but also pretty handy for moving stations from one instance of AC to another withouth having to rebuilt everything by hand

## Screenshots

grafik

grafik

grafik

grafik


frontend/components/Stations/Playlists/PlaylistGroupingTab.vue
```diff
@@ -0,0 +1,840 @@
...
+                                            <icon-ic-delete />
+                                        </button>
+                                    </div>
+
+                                    <div class="d-flex flex-column align-items-end gap-2 mb-2">
+                                        <form-group-field
+                                            :id="`consecutive-plays-${index}`"
+                                            class="col-12 col-md-6 col-xl-4 col-xxl-3"
+                                            :model-value="member.consecutive_plays"
+                                            @update:model-value="doUpdateConsecutivePlays(index, $event)"
```
Vaalyn, line 357: I didn't use a `useAppRegle` on the dynamic members here for the right side playlist group content. This list is quite different from the regular forms and creating full blown regle validation field objects for each member dynamically felt like it was a tad overengineered so I opted for a plain `model-value` / `@update:model-value` approach here for now. Is that okay with you @BusterNeece ?

Vaalyn:
Small status update:

I'm currently in the process of building out a testing suite for validating if the Scheduler & QueueBuilder behave as expected.

For this I'm working on an additional export that creates a .json file from either all or a single playlist containing all their relevant configurations and assignments (media, schedule, etc) so I can then throw them into the test suite as additional scenarios to check.

That should also make it much easier to test how the system behaves when a user reports something being weird/wrong/etc. and I'm planning to make them re-importable too so we can also check them in a station by hand if neccessary. Media that's not available on the station I'll probably just let ffmpeg generate dummy files according to the .json's media details.

I'm aiming to build this as both a unit testing suite with mocked in-memory DB interactions to have a "fast" path that could also be used in the GH pipeline as well as a functional test suite that does actual DB interactions to ensure we can also verify that the real world behaves accordingly. That way we can also catch any behaviour drift and more easily see if it comes from the DB interaction side or the pure logic parts.

I have a very rough WIP for both of those that already kinda work but definitely need a lot of polish.

BusterNeece:
@Vaalyn Excellent! I'm keeping an eye on this just to see how it's going, but let me know if my help is needed anywhere. Great work on this so far.

Vaalyn:
I have just pushed my work on the test suite, resulting fixes & adjustments.

This also fixes the previously failing tests which were partly broken by the property hooks refactor as well as by stuff from before that.

In order to be able to use unit tests with mocking to test the scheduler & queue builder in-memory I've added the [dg/bypass-finals](https://packagist.org/packages/dg/bypass-finals) dev dependency and configured it explicitly for the test suite & only for the repos since those are the only classes that I needed to mock but were final so couldn't be mocked without it.

This has become a rather large commit since I had to first built it up to work at all in the beginning and then refine it over time, fixing and adjusting parts here and there, cleanup and building up the test cases before I wanted to commit it 😅 

**Transparency note:**

I've used a local LLM to assist me with creating the test case `.json` files (`*.dump.json` & `*.scenario.json`) since that would have taken me much longer to gather otherwise (at least a month or more, depending on my available time) and also to double check my assumptions so I'm not building silently failing tests. I felt that this falls within the 100% human-coded AI policy of AzuraCast since I'm still the one building this feature / branch myself and also fully responsible for what I'm committing.

RM-FM:
@Vaalyn Great to see this progressing! 👍 I can't wait for the first rolling release that includes this.

Vaalyn:
Pushed some additional test cases to cover more of the scheduler & queue builder logic. Found & fixed a bug with scheduled overnight playlists too.

Also added a monolog TestHandler to the test harness to capture & format the debug log entries generated by the scheduler / queue builder for the assert outputs to help with understanding why a failure happened.

Parts that are still TODO for me:
- [x] See if I can expand the test harness to be able to also test requests
- [x] Schedule View
  - [x] Show groups correctly
  - [x] Find a way to better show group details there
- [x] All Playlists
  - [x] Mark playlists that are members of group/s somehow to better distinguish them
  - [x] Add list of playlist groups a playlist is a member of in a new tab when editing a playlist
- [x] Dashboard
  - [x] Show Playlist Group (maybe full path if nested) of currently playing playlist in "now playing" & "playing next"
- [x] Song Playback Timeline
  - [x] Show Playlist Group (maybe full path if nested) in addition to the playlist a track was played from
- [x] Upcoming Song Queue
  - [x] Show Playlist Group (maybe full path if nested) in addition to the playlist a track was picked from

When those are finished (and I haven't forgotten anything) it would be ready to review & test from my point of view 🤔

Vaalyn:
Schedule View adjustments are done.

Added the icons I already used in the group management tab to the schedule events to better differentiate between playlist types, also added an overlay panel on hover with more details about the playlist of a schedule event.

Also fixed a bug with the 24h time preference user setting not being applied to the schedule events time format.

grafik

grafik

Vaalyn:
Added 2 new tabs, members & memberships, to the edit modal for playlists to show these details about them and have them linked to the respective playlists so one can quickly jump to the edit modal for that playlist from here. It's currently pretty bare bones, no further details or icons here, wanted ot have it at all and probably with extend it later on if needed.

Also added a new column to the playlists overview that shows how many groups a playlist is a member of and linking to the memberships tab of its edit modal when clicking on it.

grafik

grafik

grafik

Vaalyn:
Since playlists that are part of groups are excluded from the regular playout scheduling I added a warning to the schedule view when a playlist is scheduled outside of its group (or one of the parent groups if deeper nested) so one can more easily see if a scheduled playlist would not be played due to this:

grafik

Vaalyn:
Added a new PlaylistSourceBadge component to show the playlist a song was picked from incl. the playlist group chain it was picked through (if applicable) and used it for:
- Dashboard
  - Now Playing
  - Playing Next
- Song Playback Timeline
- Upcoming Song Queue

Also added the PlaylistSourceIcon to the Playlists pages source column and added group chain verification to the test cases to make sure we check that there too.

grafik

grafik

grafik

grafik

Vaalyn:
I think I'm done with everything I've had on my TODO as well as my other notes and don't really have anything more I can think of as relevant for this PR.

Gonna tidy up my readme.md for the autodj test harness fixtures to be better understandable for others and commit that too.

I had that cobbled together mainly to help my LLM with creating those fixtures. Also I don't want to keep that just to myself since it helpe a lot with remembering what's already included 😅 

A cleaned up version of that should help others with understanding them better too I think (as well as me when I've not touched them for some time).

Vaalyn:
Added the cleaned up readme, now I've done all I could think of for this PR, thank you @RM-FM for the inspiration regarding Clock Wheel specifics 👍 

Sorry for the final size of the PR, it ended up a lot bigger than I anticipated, though ~50% of the lines is just from the test case fixture .json files.

Having that in place now makes me much more comfortable with touching the Scheduler & QueueBuilder related parts of the code though since I'm not having to risk blindly breaking things anymore so I think that should justify its size a bit 😅 

I know that reviewing this PR will be quite a big ask / task so if I can help in any way or if there is anything you'd question just tag me in a comment on it and I'll take care of it @BusterNeece

RM-FM:
@Vaalyn I think there is an issue with the `Song Playback Order` setting for playlists assigned to a Playlist Group. These playlists appear to lose the selected radio-button value for `Song Playback Order`.

Bildschirmfoto 2026-08-05 um 22 15 35

Vaalyn:
> @Vaalyn I think there is an issue with the `Song Playback Order` setting for playlists assigned to a Playlist Group. These playlists appear to lose the selected radio-button value for `Song Playback Order`.

🤔 strange, will have a look at it and check what's going on there. Might take until the weekend though.

Vaalyn:
@RM-FM the problem with the settings is fixed now.

I missed that the different sub parts of the form for the playlist types where still beeing rendered since they were only hidden via `v-show` which meant that the fields that were in both `songs` & `playlists` type playlists were clashing / confusing vue when trying to set their given values. I've switched these sub parts to `v-if` to prevent this now.

I've also added a link / button on the `# Entries` column for playlist groups that opens the edit modals members tab analogous to the `Groups` columns button.

RM-FM:
I noticed a change in the rotation behavior when not using Playlist Groups. There used to be a per-playlist feature called **“Prioritize over listener requests”** which appears to have been removed.

I don’t see any way to achieve the same behavior now, as I would be forced to use Playlist Groups and configure the allowed slots for Listener Requests there. We use this feature specifically to ensure that jingles are played at the exact slots we have planned for them, preventing a Listener Request from postponing them.

I think we should preserve the existing behavior here, as not all users (myself included) would use Playlist Groups for all of their stations.

**A pragmatic approach could also be to prevent Listener Requests from being queued if there is already a playlist configured to play every X songs for that specific “slot” in the queue.**

Vaalyn:
> I noticed a change in the rotation behavior when not using Playlist Groups. There used to be a per-playlist feature called **“Prioritize over listener requests”** which appears to have been removed.
> 
> I don’t see any way to achieve the same behavior now, as I would be forced to use Playlist Groups and configure the allowed slots for Listener Requests there. We use this feature specifically to ensure that jingles are played at the exact slots we have planned for them, preventing a Listener Request from postponing them.
> 
> I think we should preserve the existing behavior here, as not all users (myself included) would use Playlist Groups for all of their stations.
> 
> **A pragmatic approach could also be to prevent Listener Requests from being queued if there is already a playlist configured to play every X songs for that specific “slot” in the queue.**

Yes this was an intentional change, see:

grafik

The old toggle was pretty broken if used like described above.

If you'd like more control over how Requests are played out I'd go with setting the new global `Only play requests via Request Queue playlists` toggle and creating playlists of type `Request Queue` and either schedule them or set them to every x Songs / Minutes as needed.

Or setting a `Block Request Queue While Active` on a scheduled playlists schedule entry/entries as needed and create specific scheduled `Request Queue` playlist/s for that same block with the needed frequency, if any should play then.

Or if you want to only block them when the playlist is scheduled just setting the `Block Request Queue While Active` toggle on the schedule entry/entries.

This should give you all the options needed to achieve what you had previously but without the broken global blocking behaviour / making it much more explicit instead of an unintended side effect.

There shoudl be no need to use a playlist group however.

RM-FM:
> Or setting a `Block Request Queue While Active` on a scheduled playlists schedule entry/entries as needed and create specific scheduled `Request Queue` playlist/s for that same block with the needed frequency, if any should play then.

@Vaalyn Ah, I guess that was the piece I was missing ;-)

PR #8482, opened 2026-05-30
Fix issue #8481 - Refactor entities to replace direct identifier access with virtual computed properties with get hooks

### LINKED ISSUE #8481: Creating playlists causes "$station_id must not be accessed before initialization" error
### Installation Method

Docker Installation

### AzuraCast Release Channel

Rolling Release Channel

### Current AzuraCast Version

Rolling Release #62642b3 (2026-05-28 16:17)

### What happened?

When trying to create a new playlist the DoctrineEntityNormalizer is throwing an exception:
```
Typed property App\Entity\StationPlaylist::$station_id must not be accessed before initialization
at vendor/azuracast/doctrine-entity-normalizer/src/DoctrineEntityNormalizer.php:316
```

The playlist was still being created so this is most likely an issue with the normalization for the response.

### Relevant log output

```shell
Typed property App\Entity\StationPlaylist: :$station_id must not be accessed before initialization {
	"file": "/var/azuracast/www/vendor/azuracast/doctrine-entity-normalizer/src/DoctrineEntityNormalizer.php",
	"line": 316,
	"code": 0,
	"trace": [
		{
			"namespace": "",
			"short_class": "",
			"class": "",
			"type": "",
			"function": "",
			"file": "/var/azuracast/www/vendor/azuracast/doctrine-entity-normalizer/src/DoctrineEntityNormalizer.php",
			"line": 316,
			"args": []
		},
		{
			"namespace": "",
			"short_class": "ReflectionProperty",
			"class": "ReflectionProperty",
			"type": "->",
			"function": "getValue",
			"file": "/var/azuracast/www/vendor/azuracast/doctrine-entity-normalizer/src/DoctrineEntityNormalizer.php",
			"line": 316,
			"args": [
				[
					"object",
					"App\\Entity\\StationPlaylist"
				]
			]
		},
		{
			"namespace": "Azura\\Normalizer",
			"short_class": "DoctrineEntityNormalizer",
			"class": "Azura\\Normalizer\\DoctrineEntityNormalizer",
			"type": "->",
			"function": "getProperty",
			"file": "/var/azuracast/www/vendor/azuracast/doctrine-entity-normalizer/src/DoctrineEntityNormalizer.php",
			"line": 246,
			"args": [
				[
					"object",
					"ReflectionClass"
				],
				[
					"object",
					"App\\Entity\\StationPlaylist"
				],
				[
					"string",
					"station_id"
				]
			]
		},
		{
			"namespace": "Azura\\Normalizer",
			"short_class": "DoctrineEntityNormalizer",
			"class": "Azura\\Normalizer\\DoctrineEntityNormalizer",
			"type": "->",
			"function": "getAttributeValue",
			"file": "/var/azuracast/www/vendor/symfony/serializer/Normalizer/AbstractObjectNormalizer.php",
			"line": 198,
			"args": [
				[
					"object",
					"App\\Entity\\StationPlaylist"
				],
				[
					"string",
					"station_id"
				],
				[
					"null",
					null
				],
				[
					"array",
					{
						"ignored_attributes": [
							"array",
							[
								[
									"Over 9 levels deep, aborting normalization",
									"Over 9 levels deep, aborting normalization"
								]
							]
						],
						"enable_max_depth": [
							"boolean",
							true
						],
						"max_depth_handler": [
							"object",
							"Closure"
						],
						"circular_reference_handler": [
							"object",
							"Closure"
						],
						"class_metadata": [
							"object",
							"Doctrine\\ORM\\Mapping\\ClassMetadata"
						],
						"association_mappings": [
							"array",
							{
								"station": [
									"Over 9 levels deep, aborting normalization",
									"Over 9 levels deep, aborting normalization"
								]
							}
						],
						"_read_attributes": [
							"boolean",
							true
						],
						"cache_key": [
							"boolean",
							false
						],
						"circular_reference_limit_counters": [
							"array",
							{
								"4811": [
									"Over 9 levels deep, aborting normalization",
									"Over 9 levels deep, aborting normalization"
								]
							}
						]
					}
				]
			]
		},
		{
			"namespace": "Symfony\\Component\\Serializer\\Normalizer",
			"short_class": "AbstractObjectNormalizer",
			"class": "Symfony\\Component\\Serializer\\Normalizer\\AbstractObjectNormalizer",
			"type": "->",
			"function": "normalize",
			"file": "/var/azuracast/www/vendor/azuracast/doctrine-entity-normalizer/src/DoctrineEntityNormalizer.php",
			"line": 64,
			"args": [
				[
					"object",
					"App\\Entity\\StationPlaylist"
				],
				[
					"null",
					null
				],
				[
					"array",
					{
						"ignored_attributes": [
							"array",
							[
								[
									"Over 9 levels deep, aborting normalization",
									"Over 9 levels deep, aborting normalization"
								]
							]
						],
						"enable_max_depth": [
							"boolean",
							true
						],
						"max_depth_handler": [
							"object",
							"Closure"
						],
						"circular_reference_handler": [
							"object",
							"Closure"
						],
						"class_metadata": [
							"object",
							"Doctrine\\ORM\\Mapping\\ClassMetadata"
						],
						"association_mappings": [
							"array",
							{
								"station": [
									"Over 9 levels deep, aborting normalization",
									"Over 9 levels deep, aborting normalization"
								]
							}
						],
						"_read_attributes": [
							"boolean",
							true
						],
						"cache_key": [
							"boolean",
							false
						],
						"circular_reference_limit_counters": [
							"array",
							{
								"4811": [
									"Over 9 levels deep, aborting normalization",
									"Over 9 levels deep, aborting normalization"
								]
							}
						]
					}
				]
			]
		}
	]
} []
```

Vaalyn:
**Fixes issue:**
Fixes #8481 

**Proposed changes:**
This PR refactors our entities to replace the `*_id` properties that we used for directly accessing the identifiers of the entities (and relations) with virtual computed properties with get hooks.

The problem described in the issue was caused by the normalizer trying to access the playlists station_id field when normalizing the created entity for the response, but using the entity from the memory without hydrating this field since it the entity has not been initialized from doctrine itself.

I could have tried to let it hydrate those entities but since this would have just been a crutch for fixing an error caused by an implementation that we wanted to replace anyways I opted for just refactoring the TODOs instead.

Only found 2 DQL queries that used the scalar fields and adjusted them to use the IDENTITY function.

### DIFF
--- a/backend/src/Controller/Api/Stations/Files/ListAction.php
+++ b/backend/src/Controller/Api/Stations/Files/ListAction.php
@@ -176,11 +176,11 @@ final class ListAction implements SingleActionInterface
                         );
                     } elseif ('unassigned' === $special) {
                         $mediaQueryBuilder->andWhere(
-                            'sm.id NOT IN (SELECT spm2.media_id FROM App\Entity\StationPlaylistMedia spm2)'
+                            'sm.id NOT IN (SELECT IDENTITY(spm2.media) FROM App\Entity\StationPlaylistMedia spm2)'
                         );
                     } elseif (null !== $playlist) {
                         $mediaQueryBuilder->andWhere(
-                            'sm.id IN (SELECT spm2.media_id FROM App\Entity\StationPlaylistMedia spm2 '
+                            'sm.id IN (SELECT IDENTITY(spm2.media) FROM App\Entity\StationPlaylistMedia spm2 '
                             . 'WHERE spm2.playlist = :playlist)'
                         )->setParameter('playlist', $playlist);
                     }
--- a/backend/src/Entity/Listener.php
+++ b/backend/src/Entity/Listener.php
@@ -28,33 +28,33 @@ final class Listener implements
     #[ORM\JoinColumn(name: 'station_id', referencedColumnName: 'id', nullable: false, onDelete: 'CASCADE')]
     public readonly Station $station;
 
-    /* TODO Remove direct identifier access. */
-    #[ORM\Column(nullable: false, insertable: false, updatable: false)]
-    public private(set) int $station_id;
+    public int $station_id {
+        get => $this->station->id;
+    }
 
     #[ORM\ManyToOne(targetEntity: StationMount::class)]
     #[ORM\JoinColumn(name: 'mount_id', referencedColumnName: 'id', nullable: true, onDelete: 'SET NULL')]
     public readonly ?StationMount $mount;
 
-    /* TODO Remove direct identifier access. */
-    #[ORM\Column(nullable: true, insertable: false, updatable: false)]
-    public private(set) ?int $mount_id = null;
+    public ?int $mount_id {
+        get => $this->mount?->id;
+    }
 
     #[ORM\ManyToOne(targetEntity: StationRemote::class)]
     #[ORM\JoinColumn(name: 'remote_id', referencedColumnName: 'id', nullable: true, onDelete: 'SET NULL')]
     public readonly ?StationRemote $remote;
 
-    /* TODO Remove direct identifier access. */
-    #[ORM\Column(nullable: true, insertable: false, updatable: false)]
-    public private(set) ?int $remote_id = null;
+    public ?int $remote_id {
+        get => $this->remote?->id;
+    }
 
     #[ORM\ManyToOne(targetEntity: StationHlsStream::class)]
     #[ORM\JoinColumn(name: 'hls_stream_id', referencedColumnName: 'id', nullable: true, onDelete: 'SET NULL')]
     public readonly ?StationHlsStream $hls_stream;
 
-    /* TODO Remove direct identifier access. */
-    #[ORM\Column(nullable: true, insertable: false, updatable: false)]
-    public private(set) ?int $hls_stream_id = null;
+    public ?int $hls_stream_id {
+        get => $this->hls_stream?->id;
+    }
 
     #[ORM\Column]
     public int $listener_uid;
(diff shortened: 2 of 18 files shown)


BusterNeece:
@Vaalyn This does fix the issue in this case, but the reason I kept using those identifiers is because Doctrine hydrates them without hydrating the related entity, so calling them in cases where they're needed allowed me to fetch them without having to re-fetch/re-hydrate the related entity.

I think the greater "todo" attached to several of them still applies, in the sense that ideally we'd be using queries with `IDENTITY(relation)` and such things where possible to optimize.

Vaalyn:
🤔 if I didn't misunderstood your point then I think this might only be historically true (re: doctrine hydrating the ids without also hydrating the whole related entity) but in the current version of Doctrine combined with our settings, as far as I can tell, this is not really a concern anymore:

- We have configured the Doctrine Entity Manager with `$config->enableNativeLazyObjects(true);`
- When that is enabled the Doctrine ORM Proxy directly sets the id to the id fields on the lazy ghost without loading the entire entity already (via `setRawValueWithoutLazyInitialization` in the [RawValuePropertyAccessor](https://github.com/doctrine/orm/blob/3.6.x/src/Mapping/PropertyAccessors/RawValuePropertyAccessor.php#L39-L52)) and as such only loads the entity if a field is touched that is not already set (so anything but the ID) via the `loadById` in the callback
  - See: https://github.com/doctrine/orm/blob/3.6.x/src/Proxy/ProxyFactory.php#L210-L242
  - Also see `Non-Triggering Operations` on https://www.php.net/manual/en/language.oop5.lazy-objects.php
- The getter property should, in my understanding, not cause any hydration of the related entity here

(Went down this rabbit hole to fully understand it myself and wanted to share my findings 😅)

Regarding the greater TODO about using `IDENTITY(relation)`, I tried to look at all the places where we are doing queries but I haven't really found any queries apart from the ones I adjusted that would be candidates for this since in all the places I checked we also need access to either other properties too or the entity as such (for removal for example).

BusterNeece:
@Vaalyn Ah, I did not realize that the lazy-loaded proxies actually had the ID basically set on them without needing to hydrate the whole object. In that case, we really aren't losing anything doing it this way. Thank you for looking into all that!

PR #8494, opened 2026-06-05
fix(api): refresh scheduled entity after flush before serialization

souzadavi:
When creating StationPlaylist/StationStreamer records, Doctrine flush may leave DB-managed scalar fields uninitialized on the in-memory entity (e.g. StationPlaylist::$station_id).

The API immediately serializes the entity in the same request, which can trigger: Typed property App\Entity\StationPlaylist::$station_id must not be accessed before initialization.

Call EntityManager::refresh($record) right after flush in AbstractScheduledEntityController::editRecord() so the entity is fully hydrated before response serialization.

**Fixes issue:**

**Proposed changes:**


Vaalyn:
This bug has been fixed already via https://github.com/AzuraCast/AzuraCast/pull/8482

PR #8498, opened 2026-06-06
feat: per-station custom public-page domain

w3K-one:
First, I want to say — this project is genuinely impressive. The architecture, the attention to detail, the community around it.

**mETaLmuSicRaDio** has been broadcasting since **1996** — when internet radio meant RealMedia and RealAudio and you felt lucky if anyone could buffer it. We moved to WinAmp + Shoutcast when that became the combination to beat, then to **SAM Broadcaster** in 2002, pairing it with Shoutcast and Windows Media Server for years until we dropped WMS around 2008. Since then it's been SAM + Shoutcast, with Icecast mixed in at times, for nearly two more decades.

SAM Broadcaster is all but abandoned now. Finding AzuraCast — something modern, actively maintained, open, and genuinely capable — is exactly what we needed. It's the first platform in 30 years that felt like a real upgrade rather than a lateral move. We forked it as **mETaLcast**, made it ours, and got it running. And now we want to give something back.

---

### Custom Public Page Domain

This PR adds an optional **Custom Public Page Domain** field to each station (found under Station Profile → Advanced). When set, AzuraCast writes a dedicated nginx `server{}` block that serves the station's public player at the root of that domain, proxying transparently to the internal server on `127.0.0.1:6010`.

**What it does:**
- Lets a station have its own vanity domain (e.g. `radio.example.com`) that shows the public player at `/`, with no `/public/{slug}` in the URL
- The browser URL stays clean — no redirects, transparent proxy
- Static assets, API, `/listen`, `/embed`, `/history` and other sub-paths all work correctly
- SSL uses the existing ACME certificate already managed by AzuraCast

**How it works:**
- A new `custom_domain` column on the `station` table (migration included)
- `Nginx::writeCustomDomainConfiguration()` generates a `custom_domain.conf` file in the station's config directory (`/var/azuracast/stations/{slug}/config/`) — a path the `azuracast` user can write to, unlike `/etc/nginx/sites-enabled/`
- The nginx template gets a new http-level `include` for `stations/*/config/custom_domain.conf` — placed outside the main `server{}` block so full `server{}` entries are valid
- `Nginx::testConfiguration()` runs `nginx -t` before every reload so a malformed config is caught before it can affect the running server
- The domain is validated against a strict hostname regex before anything is written to disk
- The API triggers a config rewrite when the domain changes, and writes the config immediately when a domain is set at station creation

**Files changed:**
- `backend/src/Entity/Station.php` — `custom_domain` property
- `backend/src/Entity/Migration/Version20260605120000.php` — DB migration
- `backend/src/Nginx/Nginx.php` — `writeCustomDomainConfiguration()`, `testConfiguration()`
- `backend/src/Controller/Api/Admin/StationsController.php` — trigger on domain change / creation
- `backend/src/Console/Command/RestartRadioCommand.php` — write config on station restart
- `util/docker/web/nginx/azuracast.conf.tmpl` — http-level include
- `frontend/entities/ApiInterfaces.ts` — `custom_domain` on Station type
- `frontend/components/Admin/Stations/Form/form.ts` — form default
- `frontend/components/Admin/Stations/Form/ProfileForm.vue` — UI field (Advanced section)


Vaalyn:
Has this been worked on by a coding agent like Claude, Codex, etc.?

w3K-one:
**Update — two bugs fixed in the nginx generator (commit 4a5414e)**

After running this in production on our own station for several days, we found two issues in the generated `custom_domain.conf`:

---

**1. Stream mount URLs returned 404**

The original `location = /listen` is an exact-match directive — it only matches the literal URI `/listen`, not `/listen/192K.mp3` or any mount-point URL. Listeners trying to connect to any stream via the custom domain got a 404.

Fixed: replaced with `location ^~ /listen/` which proxies directly to the station's streaming server on its configured port (`$station->frontend_config->port`). The `^~` prefix strips `/listen/` so the streaming server receives the mount name as `/[mount]`.

---

**2. Public player showed "Station offline" on the custom domain**

`/api/live/nowplaying/*` (the Centrifugo SSE/WebSocket endpoint the public player uses for live Now Playing updates) was being caught by `location ^~ /api` and routed to the PHP backend on port 6010. Centrifugo is on port 6020 — the PHP backend has no stream proxy, so it returned an error and the player fell back to "Station offline".

Fixed: added `location ^~ /api/live/` with a rewrite to `/connection/uni_$1` and `proxy_pass` to port 6020, placed **before** the generic `/api` block. In nginx, the longer `^~` prefix (`/api/live/`) wins over the shorter one (`/api`), so live Now Playing requests now reach Centrifugo correctly.

---

Both issues are verified fixed in production. The rest of the PR (entity, migration, frontend, azuracast.conf.tmpl) is unchanged.

Vaalyn:
> Has this been worked on by a coding agent like Claude, Codex, etc.?

Any input to this question?

w3K-one:
> > Has this been worked on by a coding agent like Claude, Codex, etc.?
> 
> Any input to this question?

Sorry, I didn't notice this before. I did use AI to sanity‑check for bugs and improve the write‑up/wording of the PR description and comments.

backend/src/Nginx/Nginx.php
```diff
@@ -32,17 +33,149 @@ public function writeConfiguration(
...
+            return false;
+        }
+
+        $domain = $station->custom_domain;
+        $stationSlug = $station->short_name;
+
+        if (!preg_match('/^(?=.{1,253}$)([a-z0-9]([a-z0-9-]{0,61}[a-z0-9])?\.)+[a-z]{2,63}$/i', $domain)) {
+            throw new \InvalidArgumentException('Invalid custom domain: ' . $domain);
+        }
+        if (!preg_match('/^[a-z0-9_-]+$/', $stationSlug)) {
```
Vaalyn, line 78: We currently don't have a regex for validating the station slug when writing NGINX configs or in any other place that uses it.

- Why do we need it here if we don't have it at other places?
- Why this specific one?
- Why not validate it in the setter of the Station entity?

backend/src/Nginx/Nginx.php
```diff
@@ -32,17 +33,149 @@ public function writeConfiguration(
...
+                $fs->remove($customDomainConfigPath);
+                return true;
+            }
+            return false;
+        }
+
+        $domain = $station->custom_domain;
+        $stationSlug = $station->short_name;
+
+        if (!preg_match('/^(?=.{1,253}$)([a-z0-9]([a-z0-9-]{0,61}[a-z0-9])?\.)+[a-z]{2,63}$/i', $domain)) {
```
Vaalyn, line 75: This will fail on domains with umlauts like for example `schönes-radio.de`.

- Why validate here and not in the setter on the Station entity?
  - I'd also use the same kind of validation we have on the Settings::$base_url instead of a regex for this
- You could just construct an instance of `GuzzleHttp\Psr7\Uri` here and get the domain from it via `getHost`?

backend/src/Nginx/Nginx.php
```diff
@@ -32,17 +33,149 @@ public function writeConfiguration(
...
+            throw new \InvalidArgumentException('Invalid custom domain: ' . $domain);
+        }
+        if (!preg_match('/^[a-z0-9_-]+$/', $stationSlug)) {
+            throw new \InvalidArgumentException('Invalid station slug: ' . $stationSlug);
+        }
+        $stationName = str_replace(["\n", "\r", '#'], ' ', $station->name);
```
Vaalyn, line 81: Why these string replaces here?

Nowehere else do we do this when using the stations name for anything...

backend/src/Nginx/Nginx.php
```diff
@@ -32,17 +33,149 @@ public function writeConfiguration(
...
+                proxy_pass http://127.0.0.1:6020;
+            }
+            location ^~ /api {
+                include /etc/nginx/proxy_params;
+                proxy_pass http://127.0.0.1:6010;
+            }
+
+            # Stream mounts: /listen/[mount] → streaming server directly.
+            # ^~ strips the /listen/ prefix; the streaming server handles /[mount].
+            location ^~ /listen/ {
```
Vaalyn, line 141: Having the location blocks for the radio proxy duplicated in the code base feels wrong since we'd have to keep any changes to it in sync in 2 places.

Also this is missing some things that are set on the regular block from the `backend/src/Nginx/ConfigWriter.php` like for example `set \$args \$args&_ic2=1;`

Vaalyn (changes requested):
In addition to the things I commented on in the code changes:

> I did use AI to sanity‑check for bugs and improve the write‑up/wording of the PR description and comments.

After having reviewed the changes I strongly feel like you also used AI to write the code itself. Here just some of the things I noticed while looking through the changes:

- In multiple places I can find the typical em-dash in the comments
- Comments explaining stuff I wouldn't expect anyone to write like this normally
  - For example the ones in: `RestartRadioCommand`, `StationsController` or in the `Nginx` class for the `$streamingPort`...
  - explaining things that need no explaining or explained in a way that an AI would explain it to someone asking it to implement something for them
- The `→` in the nginx config
  - Anyone using an IDE would write `->` instead of the unicode arrow
- The nginx config blocks don't look like they are built on the existing ones
  - They are missing things we have on the regular ones which one would have copied when defining these domain specific ones
  - What is `/connection/uni_\$1` in the rewrite of the `/api/live/` block and where does this come from?
  - What about the missing `ssl_ecdh_curve` from the original `azuracast.conf.tmpl` server block?
  - What about the missing `add_header` definitions?
- An updated `ApiInterfaces.ts` without having updated the corresponding `web/static/openapi.yml`?
  - Feels to me like a coding agent knowing that it needs the ApiInterfaces changes in the UI but didn't care to run the commands to re-generate it
  - Adding an explanation to it that doesn't exist on the corresponding entities OA\Property::description like it wrote a new one based on the one from the entity

All in all, this doesn not give me any confidence that the PR itself was actually hand written and not built with a coding agent.

BusterNeece:
Closing this PR as I agree with @Vaalyn that there are too many red flags for LLM-assisted coding here, and the resulting product doesn't meet our quality standards for the application on the whole. We strive for a 100% human-built application with AzuraCast, even if that means more experience and work is involved.

PR #8540, opened 2026-07-26
Map .fc and .cus extensions to audio/x-mod

devloic:
### Summary

`audio/x-mod` is already in `MimeType::$processableTypes`, and `.mod` and `.stm` already map to it in `MimeTypeExtensionMap`. Two further Amiga module extensions are missing, so files using them are refused before they reach the media library:

| ext | format |
|---|---|
| `.fc` | Future Composer |
| `.cus` | Custom (UADE's `custom` replayer) |

### Why the extension map is the only place this can be fixed

Neither format is detectable by content. `finfo` returns `application/octet-stream` for both, so `MimeType::getMimeTypeFromFile()` falls through to `getMimeTypeFromPath()`, the extension map returns nothing, and `MetadataManager::read()` throws `CannotProcessMediaException` with *"MIME type ... is not processable"* on upload and on every media scan.

Adding the two entries is sufficient — the resulting MIME type is already accepted downstream, so nothing else changes.

### Scope

Two lines, no behaviour change for any existing format. `.mod`/`.stm` set the precedent; this just extends the same list.

### Context

These formats do play in practice — an external Liquidsoap decoder shelling out to `uade123` handles them at playout. The extension mapping was the only thing preventing the files from being catalogued in the first place, which is why this seemed worth sending upstream rather than carrying as a local patch.

Tested against 0.23.7: with these two entries, `.fc` and `.cus` files import and scan normally; without them they land in `unprocessable_media`.

### DIFF
--- a/backend/src/Media/MimeTypeExtensionMap.php
+++ b/backend/src/Media/MimeTypeExtensionMap.php
@@ -11,6 +11,8 @@ final class MimeTypeExtensionMap extends GeneratedExtensionToMimeTypeMap
     public const array ADDED_MIME_TYPES = [
         'mod' => 'audio/x-mod',
         'stm' => 'audio/x-mod',
+        'fc' => 'audio/x-mod',
+        'cus' => 'audio/x-mod',
         // Temporary Bugfix for incorrectly mapped aac mime type in library
         // @see https://github.com/thephpleague/mime-type-detection/pull/40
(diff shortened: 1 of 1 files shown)


BusterNeece:
@devloic Thank you for your contribution. We can't merge any PRs in where the contributor hasn't signed the CLA associated with the project. Can you look into that per the message above please?

PR #8587, opened 2026-08-29
Fix issue 8578 - Add support for `Merge playlist to play as a single track` to playlist groups

### LINKED ISSUE #8578: Add support for `Merge playlist to play as a single track` to playlist groups
Currently groups cannot have the `Merge playlist to play as a single track` set on them.

In some use-cases though this would be quite useful and needed to correctly express those playout rules. For example blocks of jingles that should play out every x tracks in a defined sequence.

We should add this option for playlist groups.

Vaalyn:
**Fixes issue:**
Fixes #8578

**Proposed changes:**
This PR implements the `Merge playlist to play as a single track` advanced playlist option for playlist groups.

Also ensured that request playlist members of groups can now be set to play up to x (or all available) requests when part of a merged group.

Additionally while building the test cases for the merged groups I encountered a bug in our queue resetting that seems to have been there for quite a while:

In the `StationPlaylistMediaRepository::resetQueue` & `StationPlaylistRepository::resetPlaylistGroupQueue` we bulk update entries in the DB via DQL, this causes the DB state and Doctrines entity map to be out of sync, basicall leaving the identitiy map stale for all entries the DQL updated and that were already in the map before the DQL. When we then call `StationPlaylistMedia::played` or `StationPlaylistGroup::played` Doctrine sees that the identity map already has `0` for the `is_queued` and registers it as an empty change set, thus not triggering a DB write.

The AutoDJ build multiple queue entries per run in one process without clearing the entitiy manager in between. Whenever a sequential or shuffle resets its queue in a build run, all entries that were selected before the reset (and were thus managed already) then went stale. The rotation would then repeat its first track or group member instead of advancing, until the next build run with a new entity manager.

The in-memory harness can't detect this since there we don't involve Doctrines entity manager, but the functional suit can detect it so I've added a regression test case for this in `group-merge-media-requeue.scenario.json`.

To fix this bug I've added a `resyncManagedEntities` method to the `backend/src/Doctrine/Repository.php` that builds a list of the stale entities and refetches them with a `Query::HINT_REFRESH` to ensure we use the current DB state.

### DIFF
--- a/backend/src/Controller/Api/Stations/PlaylistsController.php
+++ b/backend/src/Controller/Api/Stations/PlaylistsController.php
@@ -487,7 +487,14 @@ final class PlaylistsController extends AbstractScheduledEntityController
             $data['include_in_on_demand'] = false;
             $data['include_in_requests'] = false;
             $data['is_jingle'] = false;
-            $data['backend_options'] = [];
+
+            $backendOptions = (array) ($data['backend_options'] ?? []);
+            $data['backend_options'] = (
+                $source === PlaylistSources::Playlists
+                && in_array(StationPlaylist::OPTION_MERGE, $backendOptions, true)
+            )
+                ? [StationPlaylist::OPTION_MERGE]
+                : [];
 
             $type = PlaylistTypes::tryFrom($data['type'] ?? '');
             $data['type'] = ($type === PlaylistTypes::Advanced)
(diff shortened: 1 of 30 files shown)


Vaalyn:
@BusterNeece requested a review specifically for the doctrine identity map bugfix since I think that should have a second pair of eyes on it, it might be what caused some of the sporadic "my station repeated the same track" reports we had in the past 🤔

PR #8608, opened 2026-09-08
Sync runner: enforce child process timeouts so a hung now-playing worker releases its lock

adamrwarner:
## Problem

`AbstractSyncRunnerCommand::lockAndRunConsoleCommand()` sets `setTimeout()` / `setIdleTimeout()` on each child `Process`, but Symfony Process only enforces those from `checkTimeout()` (invoked by `wait()` / `getOutput()` style calls). The runner loop only calls `isRunning()`, which never checks the timeout. A child that hangs therefore runs indefinitely, `checkRunningProcesses()` keeps refreshing its lock, and every later iteration logs `Could not obtain lock for task "<station>"; skipping it.`

## Observed (production, rolling `:latest` as of 2026-09-07)

1. InnoDB deadlock between the per-station now-playing worker and the AutoDJ queue builder (both fire at track change):

```
*** (1) TRANSACTION: UPDATE station SET current_song_id = 4574 WHERE id = 1
    WAITING FOR: X lock on `station` PRIMARY
*** (2) TRANSACTION: UPDATE station_playlists SET played_at = '...' WHERE id = 5
    WAITING FOR: X lock on `station_playlists` PRIMARY
    (each holds an S FK lock on the other's parent row)
*** WE ROLL BACK TRANSACTION (2)
```
   followed by `Queue builder error: ... SQLSTATE[40001] ... 1213 Deadlock found` and `Now Playing error: The EntityManager is closed.`

2. The `azuracast:sync:nowplaying:station lounge` child then stayed alive for 30+ minutes (sleeping, no progress), the parent logged "Could not obtain lock" once per second, `/api/nowplaying/<station>` returned `404 Record not found` and `/api/nowplaying` returned `[]`, while Liquidsoap/Icecast kept streaming normally. `php-nowplaying` showed RUNNING in supervisor, so nothing restarted it. Killing the child recovered the API within seconds.

## Fix

In `checkRunningProcesses()`, call `$process->checkTimeout()` for running children. On `ProcessTimedOutException`: log an error, `stop(5)` the child, release its lock and drop it from the process list, so the next loop iteration respawns it.

This does not address the deadlock itself (a consistent lock ordering between the two writers, or a retry in the queue builder, would), but it removes the failure mode where one hung worker silently takes now-playing down until a human intervenes.

Tested on the affected instance as an overlay image: children are now killed at the configured timeout and the station lock is released.


Vaalyn:
Not very trusty to let claude generate and open a PR where it includes that it made it and then manually editing the PR to remove that mention, don't you think? 😉 

In our [contributions guidelines](https://github.com/AzuraCast/.github/blob/main/CONTRIBUTING.md#our-policy-on-generative-ai) we have a policy on generative AI that you might wanna read up on.

This PR does not give any indication that an actual human was involved in reviewing the code. It looks like claude was asked to "fix hung now-playing workers" and then instructed to create a PR for this by itself which is falling under vibe coding, which is not allowed per our policy.

grafik

Issue #8586, opened 2026-08-29
Saving media metadata destroys every file tag AzuraCast does not manage (ReplayGain, MusicBrainz, dates) — silent data loss


Neric-Radio:
### Installation Method

Docker Installation

### AzuraCast Release Channel

Stable Channel

### Current AzuraCast Version

0.23.8

### What happened?

**Describe the bug**

Saving a media record through the web editor rewrites the audio file and deletes every tag that
AzuraCast does not itself manage. On a FLAC file, 28 tags were reduced to 5 by a single save. The
user is not warned, and nothing in the UI suggests that editing a cue point rewrites the file.

The most damaging casualty is `REPLAYGAIN_TRACK_GAIN`. When "Use ReplayGain" is enabled, Liquidsoap
can no longer find the tag and recomputes the gain by decoding the whole file in a passive clock.
That starves the main clock and produces dead air for Icecast listeners (details and measurements
below).

**To Reproduce**

1. Tag a FLAC file locally with ReplayGain (foobar2000, rsgain, loudgain...).
2. Upload it to a station. Dump its tags — everything is intact:
   `ffprobe -v error -show_entries format_tags -of flat "file.flac"`
3. Open the media editor, set *Custom Cues: Cue-Out Point*, save. Change nothing else.
4. Dump the tags again.

**Expected behavior**

Tags AzuraCast does not manage should be preserved. Setting a cue point should not touch the file
at all, since cue points are stored in the database (`extra_metadata`) and reach Liquidsoap through
the API annotation, not through the file.

**Actual result**

Real measurement on one file (FLAC, 16 bit / 44.1 kHz):

| | Before edit | After one edit |
|---|---|---|
| Number of tags | 28 | 5 |
| `replaygain_track_gain` | `-7.45 dB` | gone |
| `replaygain_track_peak` | `1.031567` | gone |
| MusicBrainz IDs | 6 present | gone |
| DATE, BARCODE, LABEL, track/disc numbers | present | gone |
| new `TEXT` tag | absent | `205.5` |
| File size | 22,458,052 bytes | 22,488,708 bytes |

Survivors: `TITLE`, `ARTIST`, `ALBUM`, `GENRE`, plus a `TEXT` tag holding the cue-out value.

**Root cause (traced in the code, `stable`)**

1. `backend/src/Controller/Api/Stations/FilesController.php` line 373 calls
   `$this->mediaRepo->writeToFile($record);` unconditionally on every save — there is no check on
[...]

PR #8610, opened 2026-09-08
#8586 partially: Fix ID3v2 writes being aborted by malformed TXXX tag data

MartijnBastiaansen:
**Fixes issue:**

No existing issue — happy to open one first if you prefer. It is closely related to #8586,
which reports the same code path losing tag data; this PR fixes a different failure mode in it.

**Proposed changes:**

- Fix `Media/Metadata/Writer` handing getID3 a TXXX value in the wrong shape, which currently
  aborts the entire ID3v2 write on every MP3 — `title` and `artist` included, not just the
  extra tags.
- Skip extra tags that are null or empty, so a file no longer gets six empty TXXX frames
  stamped into it on every save.
- Add `tests/Unit/MediaMetadataWriterTest.php` covering both.

### What happens today

AzuraCast cannot write any ID3v2 tag to any MP3. Saving in the media editor reports success and
updates the database, but nothing reaches the file. This is not limited to files with cue
points set.

1. Change the title of any MP3 through **Files → edit → save**.
2. Inspect the file:

   ```
   $ ffprobe -v error -show_entries format_tags -of default=noprint_wrappers=1 \
       "/var/azuracast/stations/<station>/media/<path>.mp3"
   TAG:date=2026-09-06 12:05
   ```

   The title is unchanged.
3. Check `/var/azuracast/www_tmp/app-YYYY-MM-DD.log`:

   ```
   AzuraCast.ERROR: Cannot write metadata for file ".../example.mp3":
   ID3v2 TXXX data is not properly structured, FormatDataForID3v2() failed
   at /var/azuracast/www/backend/src/Media/Metadata/Writer.php:74
   ```

Confirmed on Rolling Release #ff2757d, Docker, across five unrelated files — with and without
cue points, and on a copy stripped of all existing tags, so it is not caused by leftover `PRIV`
frames from an external editor.

### Cause

`Writer` puts all extra tags in a hashmap under a single `text` key, and then wraps every value
in one more array level along with the known tags:

```php
$writeTags['text'] = !empty(array_intersect(['vorbiscomment', 'metaflac'], $tagFormats))
    ? implode(PHP_EOL, $metadata->getExtraTags())
    : $metadata->getExtraTags();

// ...

$tagData = array_map(function ($tagValue) {
    return [$tagValue];
}, $writeTags);
```

For MP3 that produces `'text' => [ ['amplify' => null, 'cue_in' => null, ...] ]`.

getID3 does map `text` to TXXX — `Write/ID3v2.php` has
`$ID3v2ShortFrameNameLookup[3]['text'] = 'TXXX'` — but it expects a **list** of rows with a
`description` and a `data` key:

```php
if (isset($txxx_data_array['description']) && isset($txxx_data_array['data'])) {
    $tag_data_id3v2['TXXX'][] = $txxx_data_array;
}
```

Those keys are absent, so `FormatDataForID3v2()` returns `false` and the **entire** ID3v2 write
is aborted — `TIT2` and `TPE1` included, not only the TXXX frames.

### Why this hits every MP3

`StationMedia::toMetadata()` calls `$metadata->setExtraTags($this->extra_metadata->toArray())`.
All six properties of `StationMediaMetadata` are declared `public ?float $x = null`, so they are
always initialized, and `AbstractArrayEntity::toArray()` includes every initialized public
property regardless of value. The extra tags are therefore never empty:

```php
['amplify' => null, 'cross_start_next' => null, 'cue_in' => null,
 'cue_out' => null, 'fade_in' => null, 'fade_out' => null]
```

So `if (!empty($metadata->getExtraTags()))` is always true, the TXXX branch is always taken, and
it always fails.

### Why nobody notices

`Writer` throws a `RuntimeException`, but `MetadataManager::write()` catches every `Throwable`
and only logs it. The API returns success, so a total write failure looks exactly like a
successful save.

### Tests

`tests/Unit/MediaMetadataWriterTest.php` builds a small silent MP3 in memory, so no binary
fixture is added to the repository.

- `testExtraTagsDoNotPreventKnownTagsFromBeingWritten`
- `testExtraTagsAreWrittenAsIndividualTxxxFrames`
- `testEmptyExtraTagsWriteNoTxxxFrames`

Against the current `Writer` all three fail with the production error:

```
[RuntimeException] ID3v2 TXXX data is not properly structured, FormatDataForID3v2() failed
Tests: 3, Assertions: 0, Errors: 3.
```

With the change: `OK (3 tests, 5 assertions)`.

The vorbiscomment/metaflac branch keeps its existing behaviour.

### Noted while investigating, out of scope here

- The vorbiscomment/metaflac branch has a separate problem: `implode(PHP_EOL, $extraTags)`
  discards the array keys, so cue values are written without their field names and cannot be
  read back.
- Every save also destroys tag data the writer does not manage, including the APEv2 tag carrying
  mp3gain's undo information — `DeleteTags()` runs *before* the write loop. That is #8586, and I
  have a follow-up ready for it.
- It would help a lot if `MetadataManager::write()` surfaced the failure to the caller instead of
  swallowing it. A save that writes nothing currently reports success.

---


Vaalyn:
@MartijnBastiaansen I've pushed my commits to the branch for this PR.

For the unit test, apart from extending it for my refactorings to fix the tag discarding, I've also reworked it to use our DummyMediaGenerator to generate files to test and made it run through the media repo chain to write for more real life oriented testing.

Regarding my adjustments and fixes, I'm listing them from the commits here too to better document them on this PR itself:

* Fix mtime changing on failed writes
* Only write from controller if metadata has changed
* Fix ReplayGain reading side effect from getID3 for Vorbis comments
* Remove dead tag formats from writer that we copied from the getID3 demo although we never write those
* Rework the writing of tags to avoid removing existing AC non-managed tags
* Add test cases using ffmpeg & getID3 (where needed) to verify metadata tag handling
* Fix mtime after removeAlbumArt not beeing persisted

The metadata writing should work correctly now and be protected against regressions so we do not break it aggain by accident.

Please also have a look and test this too before we merge it in, the tests say it's working fine but better safe than sorry I'd say 😄

MartijnBastiaansen:
Hey @Vaalyn ,

Was testing and checking around a bit, found 1 problem and 1 potential problem which was already there I think:

1. In one of our files there is a difference between the original file and the download back after changing some values, in one of the fields which aren't in Azuracast. Check the screenshots, first is the original file.

2. The art cover is checked against a file in the media folder. If it isn't there (yet) for any reason, the art album is removed at the write. Check StationsMediaRepository::writeToFile(), line 283 tot 286

And maybe this is more of a feature request:

1. Custom fields aren't picked up from and written to the media file. Don't know if that is on purpose, but would be nice to have it I think.

original_file

downloaded_file

Vaalyn:
@MartijnBastiaansen thanks for the testing, I've looked at and fixed the issue with the changed value, getID3's `Utils::iconv_fallback` just returned the raw data for a falsy value which caused this to become mangled then. I've added some explicit protection for this. Also corrected the `tag_encoding` setting from `UTF8` to `UTF-8` since that's what getID3 expects, no idea why we had that wrong all the time 😅 

The implicit album art removal is now also fixed.

I also added the writing of mapped custom fields when writing to the files.

Please have another look and let me know how it goes 👍

MartijnBastiaansen:
Hey @Vaalyn ,

The custom fields which are mapped (auto_assign true), but we have a couple which can't be mapped. Don't know it's possible to also save / import those easily?

Other points are working correctly :)

Vaalyn:
@MartijnBastiaansen can you give me an example for those so I can better understand what I'd need to look at?

MartijnBastiaansen:
@Vaalyn sorry there are some crucial words missing :')

The custom fields in the screenshot aren't auto_assigned and because of that they aren't read from the file or written to the file.

image

Vaalyn:
@MartijnBastiaansen I've added a new option `Custom Tag Name` to the dropdown for the tag linking of custom fields that will show an input where you can give a name under which a custom tag that does not map to one of the known tags will be saved / read from.

Let me know if that works for your use case:

grafik

MartijnBastiaansen:
@Vaalyn 

Tested it and this working fine :)

